PR #8, opened 2022-08-31
codegen: refactor templates to allow and ignore multiple tags

jenngeorge:
The default openapi-generator golang templates split code into
a file per tag. To reduce code complexity and improve naming,
we customized the templates to only expect a single tag from
the CockroachDB Cloud API's OpenAPI spec (included in this
repo as internal/spec/swagger.json).
Now we want the CockroachDB Cloud API's generated documentation
to group methods by tag, which requires including multiple tags
in its OpenAPI spec.
Instead of restricting the CockroachDB Cloud API's spec to one
tag, the following changes replace the default tag-triggered
template type with a template type that includes (and can
therefore ignore) multiple tags:

- Moved mustache templates into spec/templates.
- Moved API operation code templating from the 'API' template
type (creates one file per tag) api.mustache to the
'SupportingFile' type (creates one file) service.mustache.
- Updated spec/config.yaml to include the new service.mustache
template and mark api.mustache output files to be ignored.

The generated pkg/client code is unchanged.


erademacher:
Are there any significant changes between the old api.mustache and the new service.mustache?

jenngeorge:
> Are there any significant changes between the old api.mustache and the new service.mustache?

Not really, which is how `pkg/client/service.go` is unchanged except for an updated comment. Instead of iterating over just one api's operations, service.mustache iterates over the operations for each api. 
Here is a file with the diff from `git diff main:internal/spec/api.mustache allow-tags:internal/spec/templates/service.mustache > api_vs_service.txt` : 
[api_vs_service.txt](https://github.com/cockroachdb/cockroach-cloud-sdk-go/files/9790615/api_vs_service.txt)

erademacher:
:shipit:

jenngeorge:
TFTR!

PR #11, opened 2022-11-18
Remove AdditionalProperties from model structs as it is not used by Cockroach Cloud API

prafull01:
Removed `AdditionalProperties` field from all the model struct generated by Go SDK


erademacher:
LGTM!

PR #14, opened 2023-03-08
add support for creating multiregion serverless clusters

pjtatlow:
Pulls in the latest api spec which includes the primary_region field
on the ServerlessClusterCreateSpecification. This is necessary to
support multi-region serverless.


Makefile
```diff
@@ -17,7 +17,7 @@ generate-openapi-client:
 
 .PHONY: fetch-latest-spec
 fetch-latest-spec:
-	curl -L https://management-staging.crdb.io/assets/docs/api/latest/openapi.json > internal/spec/openapi.json
+	curl -L http://localhost:7900/assets/docs/api/latest/openapi.json > internal/spec/openapi.json
```
erademacher, line 20: Could you actually change this back to https://cockroachlabs.cloud, please? I didn't mean to change it to staging, and it shouldn't be localhost either.

erademacher (changes requested):

Makefile
pjtatlow: Oops, didn't mean to commit that 🤦🏽‍♂️

Fixed!

erademacher (approved):
LGTM!

PR #20, opened 2023-04-21
Generate int64 type for int64 instead of string

jenngeorge:
Generate int64 type for int64 instead of string

### DIFF
--- a/docs/ApiDatabase.md
+++ b/docs/ApiDatabase.md
@@ -5,7 +5,7 @@
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **Name** | **string** |  | 
-**TableCount** | Pointer to **string** |  | [optional] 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -8,7 +8,8 @@ generate-openapi-client:
 			-g go \
 			-i internal/spec/openapi.json \
 			-o internal/openapi-generator \
-			-c internal/spec/config.yaml
+			-c internal/spec/config.yaml \
+			--type-mappings=int64=int64
```
jenngeorge, line 12: I'd like this to be in https://github.com/cockroachdb/cockroach-cloud-sdk-go/blob/main/internal/spec/config.yaml instead if possible.

erademacher (approved):
Nice!!

Makefile
jenngeorge: Updated.

jenngeorge:
TFTR!

PR #22, opened 2023-04-24
Add JSON tags to unmarshal int64s as strings

erademacher:


internal/spec/templates/model_simple.mustache
```diff
@@ -19,7 +19,7 @@ type {{classname}} struct {
 {{#deprecated}}
 	// Deprecated
 {{/deprecated}}
-	{{name}} {{^required}}*{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+	{{name}} {{^required}}*{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}{{#isString}},string{{/isString}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
jenngeorge, line 22: This makes sense to me! 
I think including the tag for all `isString` is nice because it might also catch other json string cases in https://protobuf.dev/programming-guides/proto3/#json (sourced from https://github.com/grpc-ecosystem/grpc-gateway/issues/219#issuecomment-251250029)

More specifically targeting int64s seems like it might be difficult. I looked at the "request_unit_limit" part of the json model used by the templates by adding `--global-property debugModels` to the `openapi-generator generate ...` command  and did not see any `"x": true` variables we could immediately use.

PR #23, opened 2023-04-25
Map from int64 API fields to int64 rather than string

andy-kimball:
Convert int64 API fields to use integer rather than string using a jq filter on the openapi.json file. The OpenAPI uses strings to represent int64 values in order to better support Javascript. But this is awkward in Go, so generate these fields with int64 types instead. Use mustache to add a ",string" modifier to the json attribute on these fields so that the JSON unmarshaler will convert the API string to an int64 value.


filter.jq
```diff
@@ -0,0 +1,8 @@
+# Convert int64 API fields to use integer rather than string. The OpenAPI uses
+# strings to represent int64 values in order to better support Javascript. But
+# this is awkward in Go, so generate these fields with int64 types instead. Use
+# mustache to add a ",string" modifier to the json attribute on these fields so
+# that the JSON unmarshaler will convert the API string to an int64 value.
+.components.schemas.UsageLimits.properties.request_unit_limit.type = "integer" |
```
erademacher, line 6: Instead of targeting individual fields, is there a way to conditionally set the data type based on the format? For all fields whose format is int64, set the data type to integer.

andy-kimball: Yes, good idea, I made that change.

erademacher: Haha. Nice!

erademacher (approved):

PR #24, opened 2023-04-26
Add JSON tags to unmarshal int64s as strings

erademacher:
This is a proposal for handling fields where we want to control whether nil values should be marshaled as null or omitted. It's just a draft/prototype, but I welcome any suggestions for improvements or other solutions. The main method shows a simple usage example.

This feature is enabled by including a vendor extension on a field in the OpenAPI spec, which can be done with proto annotations. Alternatively, I could just do this for every optional field, but it's a little more bulky to use.

Thanks for taking a look!


pkg/client/model_serverless_cluster_update_specification.go
```diff
@@ -40,16 +40,16 @@ func NewServerlessClusterUpdateSpecification() *ServerlessClusterUpdateSpecifica
...
 		return ret
 	}
-	return *o.SpendLimit
+	return o.SpendLimit.GetValue()
 }
 
 // SetSpendLimit gets a reference to the given int32 and assigns it to the SpendLimit field.
-func (o *ServerlessClusterUpdateSpecification) SetSpendLimit(v int32) {
-	o.SpendLimit = &v
+func (o *ServerlessClusterUpdateSpecification) SetSpendLimit(v *int32, marshalNull bool) {
```
andy-kimball, line 51: Ideally, the API would "make simple things simple and complex things possible". For example, could we have an API like this instead:
```
// Set spend limit (the default Set method does not take a pointer).
spec.SetSpendLimit(1)

// No spend limit set (this would set marshalNull=true).
spec.ClearSpendLimit()
```

andy-kimball:
Do we need a new type for this particular case? What if we marked fields with an extension property that:
1. Forces the field to be a pointer.
2. Does _not_ include the `omitempty` attribute in the JSON tag.

For this particular use case (`spend_limit` + `usage_limits`), I think this is all we really want/need. The old `spend_limit` field was required, so it had to be included in every Terraform config file or `ccloud` API call. By dropping `omitempty`, we're basically preserving that behavior - we'll continue _always_ passing a `spend_limit` value, even if it's `null`.

What would we be giving up if we went down that path?

andy-kimball:
> This feature is enabled by including a vendor extension on a field in the OpenAPI spec, which can be done with proto annotations.

I looked into doing that, but then backed away. My feeling was that we shouldn't pollute the OpenAPI spec with something that's hyper-specific to Golang. So that's why in the other PR I chose to pre-process the OpenAPI json with `jq`.

internal/spec/templates/utils.mustache
```diff
@@ -3,317 +3,31 @@ package {{packageName}}
...
-// PtrFloat64 is a helper routine that returns a pointer to given float value.
-func PtrFloat64(v float64) *float64 { return &v }
-
-// PtrString is a helper routine that returns a pointer to given string value.
-func PtrString(v string) *string { return &v }
-
-// PtrTime is helper routine that returns a pointer to given Time value.
-func PtrTime(v time.Time) *time.Time { return &v }
-
-type NullableBool struct {
```
andy-kimball, line 33: Interesting that these `Nullable` classes already exist, and appear to do basically the same thing as `Optional`. How come you didn't follow the pattern here and just take advantage of generics to make a `Nullable[T]` type? By following this naming and API, you'd take advantage of existing developer experience with OpenApi-generated Go code.

pkg/client/model_serverless_cluster_update_specification.go
```diff
@@ -25,8 +25,8 @@ import (
 // ServerlessClusterUpdateSpecification struct for ServerlessClusterUpdateSpecification.
 type ServerlessClusterUpdateSpecification struct {
 	// spend_limit is the maximum monthly charge for a cluster, in US cents. We recommend using usage_limits instead, since spend_limit will be deprecated in the future.
-	SpendLimit  *int32       `json:"spend_limit,omitempty"`
-	UsageLimits *UsageLimits `json:"usage_limits,omitempty"`
+	SpendLimit  Optional[int32] `json:"spend_limit,omitempty"`
```
andy-kimball, line 28: NIT: I think we should not emit `omitEmpty` if it's an `Optional` type.

erademacher:
Thanks @andy-kimball. I appreciate the comments.

Most of the problem here boils down to making sure the _user_ has control over whether null fields are included. Your suggestions would clean this up quite a bit, but would cause us to _always_ marshal these fields, and that's not always the desired behavior. We want to allow users to update other fields without accidentally erasing existing usage limits. In that case, we'd want to omit the field from the request entirely. That's why I kept "omitempty" in the json tag.

Regarding the vendor extensions, I don't think it's that much of a problem. We're only going to use this tag on the rare field where null and zero are both valid but semantically different, which is really just these two fields right now. Golang isn't the only language where we'll see this problem, either. Using `jq` or any other post-processing just makes it harder for our devs to find and add these extensions.

internal/spec/templates/utils.mustache
erademacher: The existing types are kind of similar to what I'm doing here, but not quite. They keep track of set/unset, but don't use that state to control marshaling behavior. Plus, I don't like the name "nullable". It doesn't really communicate the whole purpose. I'm not married to "optional" either, though.

PR #25, opened 2023-04-27
Retract v1.0.1 from pkg.go.dev.

andy-kimball:
A few days ago we published v1.0.1, which got picked up by pkg.go.dev. That release caused errors when unmarshaling API values into int64 fields, and so needs to be retracted.


erademacher (approved):
Thanks for handling this.

jenngeorge (approved):
Thank you!

PR #26, opened 2023-04-30
Use default MarshalJSON behavior

andy-kimball:
The OpenAPI generator generates a custom MarshalJSON method for each model. The generated code adds each model field into a map, and then marshals the map using the default JSON marshalling code. However, this generated code doesn't respect JSON field attributes like "string", which are meant to modify marshalling behavior. This issue is preventing the SDK from round- tripping certain models: the UnmarshalJSON method is not overridden and does respect JSON field attributes, while the MarshalJSON method does not respect them. This is causing errors in the ccloud tool, which expects all fields to be round-trippable.

The fix for this problem is to stop generating custom MarshalJSON methods, so that the default JSON marshalling will run instead. It doesn't seem like the custom marshalling code is doing anything special (at least not for the models we use), so it should be safe to skip it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 # Folders
 _obj
 _test
+bin
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
+# TOOLPATH is the location of tools used during builds.
+TOOLPATH := $(abspath bin)
+
 # `brew install openapi-generator` on Mac
 # Installation doc for openapi-generator: https://openapi-generator.tech/docs/installation/
 .PHONY: generate-openapi-client
-generate-openapi-client:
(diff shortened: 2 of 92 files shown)


andy-kimball:
I ran the Terraform and ccloud tests with the generated code, and they passed. However, I don't have a good sense of how complete those tests are.

erademacher (approved):
Yeah, I can't figure out why the custom MarshalJSON was necessary in the first place. I manually tested a number of different scenarios in the TF provider and in ccloud, and it all seems fine.

jenngeorge (approved):

andy-kimball:
I did some additional testing, using `gofakeit` to generate sample data and compare the output of the default `json.Marshal` function with the generated `MarshalJSON` method. I didn't find any differences, so I think it's highly likely this is a no-op change.

PR #27, opened 2023-05-09
Add Nullable struct to control marshaling of nil values

erademacher:
Previously, there was no way to selectively omit fields from a request when their value is nil. Optional fields use the omitempty tag, so there was no way to force the SDK to set the field to "null" in the JSON request.

This change adds the Nullable wrapper. The inner value is always marshaled, even if it's nil. The Nullable itself, though, still uses omitempty, so it's omitted if nil. It's set based on the `isNullable` property, which is an OpenAPI 3.0 keyword. I'm going to introduce a followup change to set it based on an x-nullable tag when upgrading our 2.0 spec to 3.0.

There was previously a set of Nullable classes for primitive types, but it didn't quite do what we wanted it to, so I removed it in favor of a generic solution.

I added an Omit* method to Nullable, but I'm not sure it's necessary. All of the CCAPI's PATCH request objects are separate from their corresponding response objects, so it's unlikely that anyone's ever going to need to call Omit*. They can just leave the field out of the struct. There are also a few other methods from the old Nullable objects in the default template that I could add here too if people think they're helpful.

I'm leaving this as a draft for now while I work on some supporting managed-service changes that'll make the official OpenAPI spec match the one in this PR.


pkg/client/model_serverless_cluster_update_specification.go
```diff
@@ -21,8 +21,8 @@ package client
 // ServerlessClusterUpdateSpecification struct for ServerlessClusterUpdateSpecification.
 type ServerlessClusterUpdateSpecification struct {
 	// spend_limit is the maximum monthly charge for a cluster, in US cents. We recommend using usage_limits instead, since spend_limit will be deprecated in the future.
-	SpendLimit  *int32       `json:"spend_limit,omitempty"`
-	UsageLimits *UsageLimits `json:"usage_limits,omitempty"`
+	SpendLimit  *Nullable[int32] `json:"spend_limit,omitempty"`
+	UsageLimits *UsageLimits     `json:"usage_limits,omitempty"`
```
andy-kimball, line 25: Does this need to be `*Nullable[UsageLimits]`? If we didn't have the spend limit (or later remove it), then there's no way to know whether `UsageLimits=nil` means "don't set this field" vs. "unlimited".

PR #29, opened 2023-05-22
Add optional retry logic to GET requests

erademacher:
I'm not sure what's up with all the markdown changes or the package downgrade. @fantapop did you forget to check something in earlier?


README.md
```diff
@@ -4,7 +4,7 @@
 This API client was generated by the [OpenAPI Generator](https://openapi-generator.tech) project.  By using the [OpenAPI-spec](https://www.openapis.org/) from a remote server, you can easily generate an API client.
 
 - API version: 2023-04-10
-- Package version: 1.1.0
+- Package version: 1.0.0
```
fantapop, line 7: This looks like its reverting the version

fantapop: Hmm, I see, this is being generated from something that I didn't know about. I changed it manually to 1.1.0.  Do you know what its updated from?  It would be good to fix that :-)

fantapop:
> I'm not sure what's up with all the markdown changes or the package downgrade. @fantapop did you forget to check something in earlier?

🤦  This was me. I was experimenting with trying to make the output look good and apparently I had generated the version with the backticks but then decided I didn't like it better so reverted that change to the template but forgot to rebuild.  It might be nice for us to run a build with a github action and compare the output. I could add that to a flex friday list.

internal/spec/templates/client.mustache
```diff
@@ -135,7 +135,23 @@ func (c *Client) callAPI(request *http.Request) (*http.Response, error) {
...
+		resp, err = c.cfg.HTTPClient.Do(request)
+		if resp.StatusCode < http.StatusInternalServerError {
+			break
+		}
+		if c.cfg.Debug {
+			log.Printf("\nError calling API: %v. Retries remaining: %d\n", err, i)
+		}
+        if i > 1 {
+		    time.Sleep(c.cfg.RetryInterval)
+        }
```
fantapop, line 153: some weird indent here

fantapop: oh, interesting, I see this is a mustache file.  I'll make this non-blocking.

pkg/client/configuration.go
```diff
@@ -28,24 +29,28 @@ const ApiVersion = "2023-04-10"
...
 // NewConfiguration returns a new Configuration object.
 func NewConfiguration(apiKey string) *Configuration {
 	cfg := &Configuration{
-		DefaultHeader: make(map[string]string),
-		UserAgent:     "ccloud-sdk-go/1.0.0",
-		Debug:         false,
-		ServerURL:     DefaultServerURL,
-		apiKey:        apiKey,
+		DefaultHeader:     make(map[string]string),
+		UserAgent:         "ccloud-sdk-go/1.0.0",
```
fantapop, line 48: It looks like I also missed updating the UserAgent version when I tried to update to 1.1.0.  Can you update?

pkg/client/client.go
```diff
@@ -152,7 +152,23 @@ func (c *Client) callAPI(request *http.Request) (*http.Response, error) {
...
 	}
 
-	resp, err := c.cfg.HTTPClient.Do(request)
+	attempts := 1
+	if request.Method == http.MethodGet {
+		attempts += c.cfg.GetRequestRetries
+	}
+
+	for i := attempts; i > 0; i-- {
+		resp, err = c.cfg.HTTPClient.Do(request)
```
fantapop, line 161: This retry logic is pretty basic.  I wonder if this would be enough for someone who wanted to retry more than once.  I've noticed that go libraries often allow you to pass in a custom http client which could handle the retries for you.  For example, when I added retries to the metronome client, I was able to just pass in retryable_http client which has exponential backoff and some other nice features built into it: https://github.com/cockroachlabs/managed-service/blob/master/console/metronome/client.go#L1129

I'm on board with keeping this basic version but I feel like we should also allow passing in a custom client and potentially mention in the read me that "more complex retry logic can be attained by passing in a custom retry logic such as... "

internal/spec/templates/configuration.mustache
```diff
@@ -23,27 +24,31 @@ const ApiVersion = "{{.}}"
...
 cfg := &Configuration{
-DefaultHeader:    make(map[string]string),
-UserAgent:        "{{{httpUserAgent}}}{{^httpUserAgent}}ccloud-sdk-go/{{{packageVersion}}}{{/httpUserAgent}}",
-Debug:            false,
-ServerURL:        DefaultServerURL,
-apiKey:			  apiKey,
+DefaultHeader:     make(map[string]string),
+UserAgent:         "{{{httpUserAgent}}}{{^httpUserAgent}}ccloud-sdk-go/{{{packageVersion}}}{{/httpUserAgent}}",
+Debug:             false,
+GetRequestRetries: 0,
```
fantapop, line 48: defaulting to off?  I'm good with it, I just wanted to make sure its not a mistake.

README.md
```diff
@@ -4,7 +4,7 @@
 This API client was generated by the [OpenAPI Generator](https://openapi-generator.tech) project.  By using the [OpenAPI-spec](https://www.openapis.org/) from a remote server, you can easily generate an API client.
```
fantapop, line 4: It seems like we should mention the new option in the README somewhere.

PR #30, opened 2023-05-23
Fix codegen diffs

erademacher:
In a previous commit, a template was changed but the resulting changes weren't checked in. Also, the package version was updated in a generated file. Checking in the generated changes and moving the package version change to the config.


fantapop (approved):
Thanks for fixing my mess!

PR #32, opened 2023-06-12
Add maintenance windows

dasrirez:


dasrirez:
It seems that the SDK was updated based on the latest CC generated API spec, this PR was not required!

PR #34, opened 2023-06-27
add support for editing serverless regions

pjtatlow:
Because the openapi spec for editing serverless regions isn't in production yet, I manually copied this one from staging before running make. If we'd rather wait for it to be in production that would be okay though.

PR #35, opened 2023-06-28
Fix typo in docs template and regenerate

erademacher:
There was a typo that was preventing notes from rendering, resulting in just a couple lines of extra whitespace on operations that had a non-empty description. This change fixes the typo and fetches the latest production spec. That should give a better base for other changes, like #34


jenngeorge (approved):
LGTM! nice catch

erademacher:
TFTR!

PR #38, opened 2023-08-07
Handle unknown enum values gracefully

erademacher:
Previously, the SDK returned an error when it encountered an enum value from the API server that it didn't recognize. This change removes that validation, since it doesn't actually provide any benefit. The SDK still validates enums initialized by the client.


rgcase (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-cloud-sdk-go/38#-:-NbKp4E847tgzElIjvol:bg7igkv)* status: 0 of 33 files reviewed, all discussions resolved

PR #44, opened 2023-11-13
Update the SDK with the Cloud 2.0 spec in a branch

andy-kimball:
The Cloud 2.0 API spec will be released as Limited Access in the future. Until it is more widely available, we don't want to update the SDK in the main branch. Therefore, this commit will happen in the api-cloud-20 branch.


rgcase (approved):
LGTM

PR #47, opened 2023-12-15
update the Cloud 2.0 SDK to use the latest spec

andy-kimball:
Rename "shared" messages and fields back to "serverless".


andy-kimball:
NOTE: this is dependent on an open PR in the managed-service repo. It's purely mechanical, based on the openapi.json spec generated there.

rgcase (approved):
LGTM

PR #48, opened 2024-02-20
[CC-27289] Update SDK to latest CC API spec

carloruiz:
The immediate motivation for the upgrade is to leverage the new cloud-neutral private endpoint connections API in the Terraform provider. Steps to generate patch in this PR: 

```
$ make fetch-latest-spec
$ make
$ git tag v1.8.0
```


docs/ApiPrivateEndpointConnection.md
```diff
@@ -0,0 +1,108 @@
+# ApiPrivateEndpointConnection
```
rgcase, line 1: Do you know why it has these `Api*` prefixes on some of these types? Is the code generator trying to namespace the types or something?

docs/CockroachCloudAddPrivateEndpointConnectionRequest.md
```diff
@@ -0,0 +1,43 @@
+# CockroachCloudAddPrivateEndpointConnectionRequest
```
rgcase, line 1: Here too, it seems like it's namespacing the type names?

docs/ApiPrivateEndpointConnection.md
carloruiz: Yea, it's happening bc a `PrivateEndpointConnection` message also exists in `commonpb`, a dependency for `apipb` so the generator namespaces the type to disambiguate. It kinda sucks we have a dependency on commonpb, though from quick perusal, it seems to only be used for `KeysetPaginationRequest` and `SortOrder`, so we should be able to ditch the dependency relatively easily

rgcase: Ya that's pretty unfortunate. Since this is public-facing, I think we should probably try to fix it before cutting a release.

What's your timeline for this? Do you need this right away? Do you think you'd be able to move things around to avoid the collision?

fantapop (approved):
Lgtm thank you!

rgcase (approved):
LGTM thanks for everything you did to push this through.

carloruiz:
Thanks Ryan and Chris !!

I saw some questionable things (unrelated to this PR) in the generated SDK, like `AddIPAllowlist2`. Will open a Jira to track those separately.

PR #52, opened 2024-03-18
add summary / deprecation to service interface template

fantapop:
Previously, the deprecation field and docs were only reflected in the generated model types for the sdk and not on the exposed interface for the service.  Projects using the client interact with the interface. This meant that no docs or deprecation notices were being propagated to any IDE loading the client.

This change adds a commented line above each interface member which includes the operation summary.  If the code is marked as deprecated, it prepends this summary with "Deprecated:".

The openapi-generator project includes some functionality for customizing output in the mustache templates but I could find nothing for taking a block of text and converting it to a comment block which would work for godoc. For this reason, I was limited in what I was able to add to godoc for each method. Namely, I could not add any multiline comments.

The documented convention for marking a golang function as deprecated is to prefix any paragraph within the godoc with "Deprecated: " and to use that paragraph for describing why it is deprecated and how to proceed. (https://go.dev/wiki/Deprecated). Despite this, in my tests with VSCode, full support for displaying the deprecation is only reached by having a single line in the doc prefixed with "Deprecated: ".

Given both the limitation with vscode and openapi-generator's current lack of support for creating multi-line comment blocks, we rely on the somewhat non-standard but functioning workaround of prefixing the summary.. In the future, we will hopefully be able to use the full multiline description in the godoc for each method and rely on a Deprecation notice being present.

Here is an example output that we now generate in service.go:

	//
	// SCIM
	//

	// Create a group
	CreateGroup(ctx _context.Context, createGroupRequest *CreateGroupRequest) (*ScimGroup, *_nethttp.Response, error)
	// Get a group by ID
	GetGroup(ctx _context.Context, id string, options *GetGroupOptions) (*ScimGroup, *_nethttp.Response, error)
	// Deprecated: Search a group by ID (Deprecated)
	GetGroup2(ctx _context.Context, id string, getGroupRequest *GetGroupRequest) (*ScimGroup, *_nethttp.Response, error)
	// List groups

Screenshot 2024-03-18 at 1 40 02 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+
+BUG FIXES:
+
+* Updated templates to include the deprecation notice in the service interface
+  docs.
+
+## 1.8.0
+
 ## 1.7.0 (September 6, 2023)
 
 NEW FEATURES:
--- a/internal/spec/templates/service.mustache
+++ b/internal/spec/templates/service.mustache
@@ -12,7 +12,13 @@ import (
 )
 
 type Service interface {
-    {{#apiInfo}}{{#apis}}// {{{baseName}}}{{#operations}}{{#operation}}
+    {{#apiInfo}}{{#apis}}
+	//
+	// {{{baseName}}}
+	//
+
+	{{#operations}}{{#operation}}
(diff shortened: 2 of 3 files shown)


rgcase (approved):
LGTM Cool this looks great. I'm surprised we didn't already have these summary comments coming through.

fantapop:
It looks like service.go is a custom template page that isn't part of the default go client generator so its not that surprising.

PR #57, opened 2024-07-31

andy-kimball:
Get the latest Cloud 2.0 spec from managed-service. This includes support for a new "provisioned_vcpus" property on the UsageLimits object.


docs/MetricExportApi.md
```diff
@@ -437,7 +445,7 @@ Name | Type | Description  | Notes
 
 ### HTTP request headers
 
-- **Content-Type**: application/json
+- **Content-Type**: Not defined
```
rgcase, line 448: I wonder what would cause this?

README.md
```diff
@@ -3,7 +3,7 @@
 ## Overview
 This API client was generated by the [OpenAPI Generator](https://openapi-generator.tech) project.  By using the [OpenAPI-spec](https://www.openapis.org/) from a remote server, you can easily generate an API client.
 
-- API version: 2023-04-10
+- API version: development
```
rgcase, line 6: Do you know what caused this change? It seems unexpected and meaningful.

rgcase (commented):
@fantapop could you take a look at this if you get a chance?

docs/MetricExportApi.md
fantapop: I saw this.  I'm assuming I broke it somehow trying to add the content-type to each call.  I'll look into it.

README.md
andy-kimball: This is deliberate, because the Cloud 2.0 API changes are still in development. I'm planning not to merge this into `main`, but into the `api-cloud-20` branch. I'm just diff'ing with `main` because the diff with `api-cloud-20` is a mess, because branches have diverged.

fantapop (approved):
All the new changes look good to me here.

andy-kimball:
I switched the target branch to `api-cloud-20` after hard resetting it to `main`, and then squash merged.

docs/MetricExportApi.md
fantapop: I investigated this.  "Not defined" is apparently a normal value here for requests that don't have a body such as get requests.  The request in question here is a POST request but it doesn't have a body.   I did an upgrade on the library that detects these and I'm assuming there was a change such that POST requests without a body are detected.

The line here seems pretty awkward to me.  I would prefer if it just left it out if there was not specified content-type.  That's a change we could probably make to the template in a separate PR.

PR #60, opened 2024-08-29
Update changelog for v2.0.1 tag

fantapop:
Previously the v2.0.0 version was requested before the module was renamed to include /v2.  Go has cached the previous tag so a new one v2.0.1 will be added.


rgcase (approved):
LGTM

PR #61, opened 2024-09-04

fantapop:
Get the latest Cloud 2.0 spec from managed-service. This includes support for a new "provisioned_vcpus" property on the UsageLimits object as well as the "usage_type" property for serverless clusters.

Note that there are a bunch of other latest spec changes mixed in here from previous api updates and this does not rev the major version like is done in the main line.  So this remains at v1.

II was thinking it might be cleaner to just redo the api-cloud-20 branch with a single commit.  I started doing that in fantapop/api-cloud-20 but I'm having trouble due to the rev of the major version.   Since these branches are just going to disappear when the cloud 2.0 spec merges into the latest spec, I'm not actually concerned with just adding it all in a single commit on top of the old one.   Let me know your thoughts @andy-kimball


fantapop:
closing this one because the api doesn't include all the necessary bits yet for passing upgrade_type on create.

PR #63, opened 2024-09-11

fantapop:
I've branched off the latest main branch and then pulled the current cloud 2.0 spec.  My plan is to force push this back overtop of the `api-cloud-20` branch.

Because it was branched from main, this branch also has its major version revved to v2.

It contains Alyshan's latest changes to add upgrade_type to both the create and update specs for serverless clusters.


andy-kimball (approved):
LGTM

andy-kimball:
Reminder to make sure to update the merge base to `cloud-20` rather than `main` before you merge. You can do that by clicking on `main` at the top of the page, e.g. "fantapop wants to merge 2 commits into `main`".

PR #64, opened 2024-09-11
fix validate and install paths for v2 rev

fantapop:
It seems that there were a couple more things that needed doing for the v2 rev I did a couple weeks back.


rgcase (approved):
LGTM

PR #65, opened 2024-09-18
Update to latest spec to get cloud 2.0 changes and tag for v3

fantapop:
Note: please look over the Changelog and let me know if there is anything obvious jumping out that I may have missed there.

I separated the major version rev out into a separate commit.

This updates the spec and generates the SDK.  I've combed through the
generated changes and listed them in the changelog.  There are breaking changes
to the SDK so I'll be revving the version to 3 in a later commit. Here are the
updates:

* RequestUnitLimit and StorageMibLimit inside UsageLimits became optional and
  have hence been updated to use pointers in the UsageLimits struct. This also
  changed the NewUsageLimits constructor which now has no required arguments.

* Added UpgradeType to Add ServerlessClusterConfig. Technically a breaking
  change to the SDK because new required fields are added to the object
  constructor.

* SpendLimit was removed as an optional value in ServerlessClusterConfig.
  UsageLimits should be used instead.

* PlanType enum now consists of the new plan types: "BASIC", "STANDARD" and
  "ADVANCED".

* Add new fields to the ClusterMajorVersion struct. Technically a breaking
  change to the SDK because new required fields are added to the object
  constructor.

* ProvisionedVirtualCpus was added to UsageLimits


rgcase (approved):
LGTM

PR #66, opened 2024-10-02
update cockroach-cloud-sdk-go to v4

dcrosta:
Imported latest OpenAPI definitions and regenerated code. See CHANGELOG.md for details on what has changed.


README.md
```diff
@@ -13,13 +13,13 @@ For more information, please visit [https://support.cockroachlabs.com](https://s
 Install the package:
```
dcrosta, line 13: @fantapop should L6/L7 be updated here? (Github won't let me comment on those lines as they're outside the diff...)

fantapop: hmm, yeah I wonder why that's not updated.  That file is generated from this template.  Here is the line:

https://github.com/cockroachdb/cockroach-cloud-sdk-go/blob/main/internal/spec/templates/README.mustache#L7

fantapop: It looks like this needs to be updated: https://github.com/cockroachdb/cockroach-cloud-sdk-go/blob/e2fef1f9062095e3e71531634ea65145e3943557/internal/spec/config.yaml#L2

internal/openapi-generator/api/openapi.yaml
```diff
@@ -9418,9 +9420,8 @@ components:
         in the type URL, for example "foo.bar.com/x/y.z" will yield type
         name "y.z".
 
-
         JSON
-
+        ====
```
fantapop, line 9424: Hmm, this is surprising.  There wasn't a merge conflict here was there?

fantapop (approved):
These look good.  I did have one question about the ==== that was added as part of the generation.

internal/openapi-generator/api/openapi.yaml
dcrosta: That is in and seems to come from openapi.json. That big blob of text seems to come from the protobuf description of `Any`, if I'm reading it right -- probably a change in protobuf library version introduced this small docs change? In any event, I don't think it's due to a merge conflict or anything like that.

PR #68, opened 2024-10-29
client: support JWT auth

pritesh-lahoti:
This change enables the SDK to access CC APIs via JWT authentication.
This will be used by the Terraform Provider, as some of our customers
have requested for using JWT for provisioning resources through Terraform,
instead of API Keys.

This change also updates the deprecated jq image.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   jq:
-    image: stedolan/jq
+    image: ghcr.io/jqlang/jq:1.7.1
     volumes:
       - ./:/go/src/github.com/cockroachdb/cockroach-cloud-sdk-go
     working_dir: /go/src/github.com/cockroachdb/cockroach-cloud-sdk-go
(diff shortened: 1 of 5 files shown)


fantapop (approved):
This all looks good to me. One thing I wanted to mention is that we've been trying to be more strict about revving the major during breaking changes. So you'd need to do that before tagging. One option to avoid is to add another constructor. Something like NewConfigurationWithOptions. But I'm okay with either option. 

Thanks for your work here!

pritesh-lahoti:
> This all looks good to me. One thing I wanted to mention is that we've been trying to be more strict about revving the major during breaking changes. So you'd need to do that before tagging. One option to avoid is to add another constructor. Something like NewConfigurationWithOptions. But I'm okay with either option.
> 
> Thanks for your work here!

Thank you for the review, @fantapop . I am not sure if you hinted for any action from my side here.
The `NewConfiguration()` is backwards-compatible due to the variadic options and should not break any existing dependencies (using API Key auth). Do let me know if I'm missing something.

fantapop:
> 
> Thank you for the review, @fantapop . I am not sure if you hinted for any action from my side here. The `NewConfiguration()` is backwards-compatible due to the variadic options and should not break any existing dependencies (using API Key auth). Do let me know if I'm missing something.

🤦   Yep,  I missed that it was variadic.  Looks all good to me.  Thanks

PR #69, opened 2024-11-12
Update to latest spec for 5.0.0 changes

fantapop:
Changes include:
* support for JWT auth
* support cidr_range configuration for GCP Advanced tier clusters


pritesh-lahoti (approved):
Thank you!

fantapop:
I ended up making a change to mark the cidr_range as always present which generated a breaking change in the generated client.  I've updated the tag to be 5.0.0 instead of 4.2.0.  Additionally I added a commit to remove the generated constructors which were causing all the breaking changes so hopefully this happens less in the future.   Due to the difference in this PR, I re-requested review.

fantapop:
I'm going to update this again. I  realized we do want some of those constructors.  Specifically any constructor used as a Request object would be good to keep because changes in required parameters in those do actually indicate a breaking change.  I'm unclear if we can easily differentiate between models that aren't used in a request (such as client.Cluster) vs. those that are.  For now I'm going to revert the removal of the constructors.

pritesh-lahoti (approved):

PR #70, opened 2025-01-22
Update to latest spec to 5.1.0

davidwding:


fantapop (commented):
It looks like there were some changes in before that didn't have the SDK generated for them yet.  I think it's just the removal of the deprecated api oidc config endpoints.   When possible, I try separate the churn commit from the commit I'm personally trying to add but its not required.  You could still do this because the PCR changes have not gone to production so you could use the live json spec pretty easily.  

A couple other notes:
1. there is a CHANGELOG.md file you should update with all the changes (including the churn changes)
2. I don't think it's worth revving the major version for the removal of the api oidc config endpoints because it was private preview that only one customer was on.

fantapop (approved):
nice, this looks good to me!

davidwding:
TFTR!

PR #73, opened 2025-04-07
Update to latest spec and tag for 6.0.0

linhcrl:
This update is technically a breaking change because the CreateUser endpoint has a new response type, CreateUserResponse.

Changes include:
* updated payload requirements for SCIM endpoints
* new response type for CreateUserEndpoint
* new labels field in Cluster and FolderResource models
* new labels field in Cluster/Folder creation/update endpoints to manage labels
* GetClusterDisruptionInfo and UpdateClusterDisruption endpoints added
* updates to GetUsers endpoint
  * StartIndex and Count params in request type
  * StartIndex and ItemsPerPage in response type)


README.md
```diff
@@ -4,7 +4,7 @@
 This API client was generated by the [OpenAPI Generator](https://openapi-generator.tech) project.  By using the [OpenAPI-spec](https://www.openapis.org/) from a remote server, you can easily generate an API client.
 
 - API version: 2024-09-16
```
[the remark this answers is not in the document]
fantapop: This is the major version string of the api.  We only rev this when we're making a backwards incompatible api change.   I think here we just show the current one.  No need to update.

linhcrl:
Due to the SCIM changes, quite a bit of files were affected by this update. Most of them are just marking fields as required though.

pritesh-lahoti (approved):
Thank you, @linhcrl !

README.md
```diff
@@ -4,7 +4,7 @@
...
-- Package version: 5.1.0
+- Package version: 6.0.0
 - Build package: org.openapitools.codegen.languages.GoClientCodegen
 For more information, please visit [https://support.cockroachlabs.com](https://support.cockroachlabs.com)
```
fantapop, line 10: There are some required changes a few lines down from here.  Replace v5 with v6 in 2 spots.

fantapop: Also there is a template for the README where they need to be updated as well.   see here: internal/spec/templates/README.mustache

fantapop (changes requested):
There were a few little things I found that should be updated as part of the major version rev.  They are hard to catch.  I always look at previous commits to get them all.   One I didn't mention anywhere is that there is a `main.go` file which can load up the library.  It should be updated to the v6 version of the library.   You can run `make validate` to run it and make sure everything is compiling.

linhcrl:
Updating the major version to v6 in the import path within `main.go` is throwing an error. What do I need to do for it to recognize it as a valid path?

```
	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
```

README.md
linhcrl: done

fantapop:
The reason is because go.mod needs also be updated for the new version.  

Add this:


```
diff --git a/go.mod b/go.mod
index e79ebd7..f09f465 100644
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module github.com/cockroachdb/cockroach-cloud-sdk-go/v5
+module github.com/cockroachdb/cockroach-cloud-sdk-go/v6

 go 1.17
```

> Updating the major version to v6 in the import path within `main.go` is throwing an error (assuming because it's not published yet). What do I need to do for it to recognize it as a valid path? Or will this only work once v6 is published?
> 
> ```
> 	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
> ```

linhcrl:
> The reason is because go.mod needs also be updated for the new version.
> 
> Add this:
> 
> ```
> diff --git a/go.mod b/go.mod
> index e79ebd7..f09f465 100644
> --- a/go.mod
> +++ b/go.mod
> @@ -1,3 +1,3 @@
> -module github.com/cockroachdb/cockroach-cloud-sdk-go/v5
> +module github.com/cockroachdb/cockroach-cloud-sdk-go/v6
> 
>  go 1.17
> ```
> 
> > Updating the major version to v6 in the import path within `main.go` is throwing an error (assuming because it's not published yet). What do I need to do for it to recognize it as a valid path? Or will this only work once v6 is published?
> > ```
> > 	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
> > ```

Awesome, that worked! Thank you!

fantapop (approved):
nice. LGTM!

PR #81, opened 2025-09-08
Update latest spec to 6.5.0

mwstobo:
NEW FEATURES:

* Added new endpoint: `GetEgressPrivateEndpoint` in LIMITED ACCESS.


mwstobo:
> There are a couple additional spots the tag should be updated. If you search for 6.0.0 in your IDE you can see them. This repo could do for some linting or automating in that regard.

Ty, I've update these version references as well!

PR #87, opened 2026-03-23
[CNSL-1939] Add changelog validation to PR checks

linhcrl:
Calls the reusable workflow from cockroachdb/actions to validate CHANGELOG.md format on PRs against main and pending-deploy-* branches. The workflow fails if the changelog is invalid and posts a comment indicating whether breaking changes are detected (entries prefixed with "Breaking Change:").

Co-Authored-By: roachdev-claude <roachdev-claude-bot@cockroachlabs.com>

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,14 @@
+name: Changelog Check
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'pending-deploy-*'
+
+jobs:
+  changelog:
+    uses: cockroachdb/actions/.github/workflows/pr-changelog-check.yml@v0
+    with:
+      base-ref: ${{ github.event.pull_request.base.ref }}
+      pr-number: ${{ github.event.pull_request.number }}


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,14 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'pending-deploy-*'
+
+jobs:
+  changelog:
+    uses: cockroachdb/actions/.github/workflows/pr-changelog-check.yml@v0.2.0
```
fantapop: @linhcrl should we go ahead and tag this?  FWIW, I think we could use a major version tag here.  But either way.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,14 @@
+name: Changelog Check
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'pending-deploy-*'
```
fantapop, line 7: do we need something to run it again on PR updates?  Or will this cover it?

.github/workflows/changelog.yml
linhcrl: updated to major version. will tag the actions repo on monday

.github/workflows/changelog.yml
linhcrl: By default, the pull_request event triggers on these activity types:                                                                                                        
  - opened                                                                                                                
  - synchronize                                                                                                       
  - reopened 

So we should be good

fantapop (approved):
This looks good to me

PR #95, opened 2026-03-29
[CNSL-1935] Add automated release PR workflow

linhcrl:
Added GitHub Actions workflow that automatically creates release PRs when automation/pending-deploy-* branches are merged to main. The workflow updates go.mod and main.go for major version changes, updates config.yaml with the new version, and regenerates the OpenAPI client to automatically update all generated files (README.md, docs/README.md, pkg/client/configuration.go, and other generated code).

Example PR: https://github.com/cockroachlabs/cockroach-cloud-sdk-go-automation-testing/pull/185

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -1,0 +1,33 @@
+name: Create Release PR
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: write  # Push to fork and create branch
+  pull-requests: write  # Create release PR
+
+jobs:
+  create-release-pr:
+    # Run on merged PRs with automation/pending-deploy- prefix, or manual workflow_dispatch
+    if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.merged && startsWith(github.event.pull_request.head.ref, 'automation/pending-deploy-'))
+    uses: cockroachdb/actions/.github/workflows/create-release-pr.yml@v0
+    with:
+      fork_owner: crl-gh-actions-pr-bot
+      fork_repo: cockroach-cloud-sdk-go
+      build_script: scripts/update_version.sh
+      files_to_commit: |
+        go.mod
+        main.go
+        internal/spec/config.yaml
+        README.md
+        docs/README.md
+        pkg/client/configuration.go
+      allow_fork_force_sync: true
+    secrets:
+      fork_push_token: ${{ secrets.FORK_PUSH_TOKEN }}
+      pr_create_token: ${{ secrets.GITHUB_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Automated release workflow that creates release PRs when `automation/pending-deploy-*`
+  branches are merged to main, updating the version number in all relevant files
+
 ## [7.1.0] - 2026-04-14
 
 ### Added
--- a/scripts/lib/actions-helpers.sh
+++ b/scripts/lib/actions-helpers.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+# Logging functions for GitHub Actions workflows
+
+# Output an informational message to stdout
+log_info() {
+  local message="$1"
+  echo "$message"
+}
+
+# Output an error message using GitHub Actions workflow command format
+# https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message
+log_error() {
+  local message="$1"
+  echo "::error::$message" >&2
+}
+
+# Output a warning message using GitHub Actions workflow command format
+log_warning() {
+  local message="$1"
+  echo "::warning::$message" >&2
+}
+
+# Output a notice message using GitHub Actions workflow command format
+log_notice() {
+  local message="$1"
+  echo "::notice::$message"
+}
+
+# Write a single-line output: set_output key value
+set_output() {
+  echo "$1=$2" >> "${GITHUB_OUTPUT:-/dev/null}"
+}
--- a/scripts/lib/validation.sh
+++ b/scripts/lib/validation.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Validation helper functions
+
+# Check if required commands are available in PATH
+# Usage: check_required_commands "cmd1" "cmd2" "cmd3"
+check_required_commands() {
+  local missing_commands=()
+
+  for cmd in "$@"; do
+    if ! command -v "$cmd" >/dev/null 2>&1; then
+      missing_commands+=("$cmd")
+    fi
+  done
+
+  if [[ ${#missing_commands[@]} -gt 0 ]]; then
+    log_error "Required command(s) not found in PATH: ${missing_commands[*]}"
+    return 1
+  fi
+
+  return 0
+}
--- a/scripts/update_version.sh
+++ b/scripts/update_version.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+# This script is called by the reusable workflow after CHANGELOG.md has been updated.
+# The VERSION environment variable is set by the reusable workflow.
+# This script should only update version numbers in files other than CHANGELOG.md.
+
+# Get the script directory
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Source helper functions
+source "$SCRIPT_DIR/lib/actions-helpers.sh"
+source "$SCRIPT_DIR/lib/validation.sh"
+
+# Check required commands are available
+check_required_commands cut sed make
+
+# Check if VERSION environment variable is set
+if [ -z "$VERSION" ]; then
+    log_error "VERSION environment variable is required"
+    log_error "This script is intended to be called by the create-release-pr workflow"
+    exit 1
+fi
+
+log_info "Updating version to $VERSION in config and documentation files..."
+
+# Extract major version from VERSION (e.g., 7.0.0 -> 7)
+MAJOR_VERSION=$(echo "$VERSION" | cut --delimiter=. --fields=1)
+
+# Update go.mod module path to match major version
+log_info "Updating go.mod..."
+sed --regexp-extended "s|(github.com/cockroachdb/cockroach-cloud-sdk-go)/v[0-9]+|\1/v${MAJOR_VERSION}|" go.mod > go.mod.tmp && mv go.mod.tmp go.mod
+
+# Update main.go import path to match major version
+log_info "Updating main.go..."
+sed --regexp-extended "s|(github.com/cockroachdb/cockroach-cloud-sdk-go)/v[0-9]+|\1/v${MAJOR_VERSION}|" main.go > main.go.tmp && mv main.go.tmp main.go
+
+# Update internal/spec/config.yaml
+log_info "Updating internal/spec/config.yaml..."
+sed "s/^packageVersion:.*/packageVersion: $VERSION/" internal/spec/config.yaml > internal/spec/config.yaml.tmp && mv internal/spec/config.yaml.tmp internal/spec/config.yaml
+
+# Generate OpenAPI client to update README files
+log_info "Generating OpenAPI client..."
+# Note: sudo is required because docker containers run as root by default,
+# creating files owned by root. We then chown them back to the current user.
+sudo make generate-openapi-client
+sudo chown --recursive "$(id --user):$(id --group)" internal/openapi-generator/ pkg/client/ docs/ README.md
+
+log_info "Version updated to $VERSION"
+
+# Validate the updated code
+log_info "Running validation..."
+make validate


linhcrl:
Tested this locally and it worked as expected

.github/workflows/finalize-release.yml
```diff
@@ -0,0 +1,76 @@
...
+            echo "No changes to commit"
+            exit 1
+          fi
+
+          # Extract changelog entries for this version
+          changelog_body=$(awk '
+            /^## \[${{ steps.extract-version.outputs.next_version }}\]/ {flag=1; next}
+            /^## \[/ {flag=0}
+            flag
+          ' CHANGELOG.md | sed '/^$/d')
```
fantapop: what do you think about returning the unreleased change log info from the extract-version action?

.github/workflows/finalize-release.yml
```diff
@@ -0,0 +1,76 @@
...
+          - Updates version to ${{ steps.extract-version.outputs.next_version }}
+          - Updates CHANGELOG.md, internal/spec/config.yaml, and README.md"
+
+          gh pr create \
+            --title "Release v${{ steps.extract-version.outputs.next_version }}" \
+            --body "$pr_body" \
+            --base main \
+            --head "update-version-${{ steps.extract-version.outputs.next_version }}"
+
+          gh pr merge "update-version-${{ steps.extract-version.outputs.next_version }}" --squash --delete-branch
```
fantapop: This kind of looks like it would merge it right now...  Is it not doing that?

.github/workflows/finalize-release.yml
```diff
@@ -0,0 +1,76 @@
...
+          git push origin "update-version-${{ steps.extract-version.outputs.next_version }}"
+
+      - name: Create Pull Request and enable auto-merge
+        if: steps.extract-version.outputs.has_unreleased == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          pr_body="## Summary
+          - Updates version to ${{ steps.extract-version.outputs.next_version }}
+          - Updates CHANGELOG.md, internal/spec/config.yaml, and README.md"
```
fantapop: It seems kind of weird to be hard coding these here.  They could get out of date and might be easy to forget about.  It's probably not important for the reviewer to seem them so WDYT about just not listing those.    Also, aren't there other places?  I think some are generated from the build but there may be more places for major version changes as well.  Such as go.mod.

.github/workflows/finalize-release.yml
```diff
@@ -0,0 +1,76 @@
...
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Create branch and update version
+        if: steps.extract-version.outputs.has_unreleased == 'true'
+        run: |
+          git checkout -b "update-version-${{ steps.extract-version.outputs.next_version }}"
+          bash internal/update-version.sh "${{ steps.extract-version.outputs.next_version }}"
+
+      - name: Commit and push changes
+        if: steps.extract-version.outputs.has_unreleased == 'true'
```
fantapop: Since these all have the same conditional, is there a reason to keep them as separate steps?

internal/update-version.sh
```diff
@@ -0,0 +1,73 @@
...
+if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    echo "Error: Version must be in semver format (e.g., 6.11.0)"
+    exit 1
+fi
+
+# Helper function to set output
+set_output() {
+    echo "has_unreleased=$1"
+    [ -n "$GITHUB_OUTPUT" ] && echo "has_unreleased=$1" >> "$GITHUB_OUTPUT"
+}
```
fantapop: It seems like we should keep github action specific stuff out of the internal script for bumping the version.

internal/update-version.sh
```diff
@@ -0,0 +1,73 @@
...
+' CHANGELOG.md > CHANGELOG.md.tmp && mv CHANGELOG.md.tmp CHANGELOG.md
+
+# Update internal/spec/config.yaml
+echo "Updating internal/spec/config.yaml..."
+sed "s/^packageVersion:.*/packageVersion: $VERSION/" internal/spec/config.yaml > internal/spec/config.yaml.tmp && mv internal/spec/config.yaml.tmp internal/spec/config.yaml
+
+# Update README.md
+echo "Updating README.md..."
+sed "s/^- Package version:.*/- Package version: $VERSION/" README.md > README.md.tmp && mv README.md.tmp README.md
+
```
fantapop: I think there is some kind of generation step required for the sdk.  For this one there should be no code changes but I thought there might be some templates that included the version that get create. 

Also, we should run `make validate` somewhere.

.github/workflows/finalize-release.yml
linhcrl: Yes it is set to merge.

Ideally it would open the PR, wait for CI to pass,  then auto-merge. However there isn't any meaningful CI checks (only the CHANGELOG linting) right now and merge would fail on conflicts, so it felt safe to do. But now that I'm thinking out loud, it's not a great idea.

Initially, I just had it so that it enables auto-merge. However, when I ran the workflow, it opened the PR, did not enable auto-merge, and did not run the linter. It just opened an "inactive" PR. I couldn't get this to work as expected.

I remember you mentioned something about auto-merge not working in some cases. Is this what you were talking about?

Should it just open the PR and wait for manual merge instead?

.github/workflows/finalize-release.yml
linhcrl: yep, I made that change in my reusable workflow: https://github.com/cockroachdb/actions/pull/20/

.github/workflows/finalize-release.yml
linhcrl: Yep I removed the specific files in my reusable workflow. And you're right, seems like I missed some. Will do an audit using more older commits. I forgot to check a major update example and some of the minor/patch examples were missing certain files that should have been updated

internal/update-version.sh
linhcrl: Can you elaborate on this comment?

.github/workflows/finalize-release.yml
linhcrl: The merging of this is equivalent to doing a release so for now, I will remove the auto merge and require human sign off/merge - at least until we verify the flow works as expected/we are back from PTO

.github/workflows/finalize-release.yml
linhcrl: Updated to not list specific files in the log and updated the script to cover all the files that need to be updated

.github/workflows/finalize-release.yml
linhcrl: Resolving this as it's no longer relevant with the reusable workflow

internal/update_version.sh
```diff
@@ -0,0 +1,41 @@
...
+set -e
+
+# This script is called by the reusable workflow after CHANGELOG.md has been updated.
+# The VERSION environment variable is set by the reusable workflow.
+# This script should only update version numbers in files other than CHANGELOG.md.
+
+# Check if VERSION environment variable is set
+if [ -z "$VERSION" ]; then
+    echo "Error: VERSION environment variable must be set"
+    echo "This script is intended to be called by the create-release-pr workflow"
```
fantapop: should we print to stderr?

internal/update-version.sh
fantapop: I looked into this.  If you run `make generate-openapi-client`, it builds the docs as well.  Part of that is copying over the README which contains the version strings.   We should probably separate the docs generation into a separate command and run it from the script you've added. 

Also, since we're doing manual modification on go files and the .mod file, Please run `make validate` from your script as well.  It ensure the package compiles.  I think you'll need go installed.

fantapop:
thanks for including the sample images.  I noticed that the Changelog entries are extremely long. I think we'll want to experiment with the prompt we're using to build those.

linhcrl:
> thanks for including the sample images. I noticed that the Changelog entries are extremely long. I think we'll want to experiment with the prompt we're using to build those.

Please note this workflow does not generate any new changelog entries. This workflow simply takes the changelog entries from the unreleased section and puts them in the commit message and PR description for visibility (to recap what is going to be released). That is what you're seeing in the images I included.

The long changelog entry you're referring to is a changelog entry for another workflow I added earlier. It's not LLM generated.

**The only thing the workflow in this PR is responsible for is updating the SDK version in the CHANGELOG and all other files that need the version number updated in preparation for the release.**

internal/update-version.sh
linhcrl: To clarify, this workflow is only meant to update the version number in all relevant files in preparation for a new sdk release. By the time we get to this workflow, all the code has been generated and merged into main.

linhcrl: Added `make validate` to the end of the update script

fantapop:
> > thanks for including the sample images. I noticed that the Changelog entries are extremely long. I think we'll want to experiment with the prompt we're using to build those.
> 
> Please note this workflow does not generate any new changelog entries. This workflow simply takes the changelog entries from the unreleased section and puts them in the commit message and PR description for visibility (to recap what is going to be released). That is what you're seeing in the images I included.
> 
> The long changelog entry you're referring to is a changelog entry for another workflow I added earlier. It's not LLM generated.
> 
> **The only thing the workflow in this PR is responsible for is updating the SDK version in the CHANGELOG and all other files that need the version number updated in preparation for the release.**

yeah makes sense. I didn't think it was coming from this PR.  It just looked long and agent generated.  You hand crafted it though?

scripts/build/update_version.sh
```diff
@@ -0,0 +1,51 @@
...
+log_info "Updating internal/spec/config.yaml..."
+sed "s/^packageVersion:.*/packageVersion: $VERSION/" internal/spec/config.yaml > internal/spec/config.yaml.tmp && mv internal/spec/config.yaml.tmp internal/spec/config.yaml
+
+# Update README.md
+log_info "Updating README.md..."
+sed "s/^- Package version:.*/- Package version: $VERSION/" README.md > README.md.tmp && mv README.md.tmp README.md
+
+# Update docs/README.md
+log_info "Updating docs/README.md..."
+sed "s/^- Package version:.*/- Package version: $VERSION/" docs/README.md > docs/README.md.tmp && mv docs/README.md.tmp docs/README.md
```
fantapop: Rather than duplicating the logic of copying the docs read me.  I think we should create a new Makefile target like `make docs` and then call it from both the existing makefile target that uses it and here.

linhcrl:
> 
> yeah makes sense. I didn't think it was coming from this PR. It just looked long and agent generated. You hand crafted it though?

I did use claude to write it. It's the changelog entry for [this PR](https://github.com/cockroachdb/cockroach-cloud-sdk-go/pull/108). I'was just trying to say it's not autogenerated by a workflow based on any prompt. I can update the relevant PR to make the entry shorter but it's a one off

scripts/build/update_version.sh
linhcrl: Ah I understand now (I think)! I added a makefile target but please note I called it `sync-readme`. `docs` felt a bit too broad as there are many other docs files that don't get touched as part of this. Additionally, `make docs` sounds like we are copying README into docs/README, but it's the other way around. Let me know what you think!

internal/update-version.sh
linhcrl: Added `make sync-readme` to copy docs/README to README

linhcrl:
@fantapop I've updated the reusable action to remove changelog entries from the commit message body.

**Why:** Commit message lines should be at most 72 characters, but changelog entries don't follow this rule. It would be unnecessarily complex to convert the changelog entries to wrap at 72 characters for the commit message. Since the individual commits already describe the changes and the changelog entries are visible in the PR description, the simpler approach is to use a hardcoded message format.

Here's the PR: https://github.com/cockroachdb/actions/pull/27

scripts/build/update_version.sh
fantapop: oh I see what's going on now.  In the normal workflow, we update the `internal/spec/config.yaml` file and then run the client generation step (make generate-openapi-client).  That step creates the client and docs which includes a README.  We then shift some of the directories around including copying the README from the docs/README location to README.  Should we actually just run the `make generate-openapi-client` change?  After understanding this better, i see that pulling the moving of the README to a separate makefile target seems silly.

fantapop:
> > yeah makes sense. I didn't think it was coming from this PR. It just looked long and agent generated. You hand crafted it though?
> 
> I did use claude to write it. It's the changelog entry for [this PR](https://github.com/cockroachdb/cockroach-cloud-sdk-go/pull/108). I'was just trying to say it's not autogenerated by a workflow based on any prompt. I can update the relevant PR to make the entry shorter but it's a one off

ah I see.  makes sense.  we are planning to generate the changelog entry from claude though right?  We should consider asking it to keep it brief.

scripts/build/update_version.sh
linhcrl: Updated to run `make generate-openapi-client`. With this change, we no longer need to update `README`, `docs/README`, and `pkg/client/configuration.go` manually. Thanks for the suggestion

internal/update-version.sh
linhcrl: Updated the shell script to run `make generate-openapi-client`, which should take care of this (removed `make sync-readme`)

scripts/lib/logging.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+# Logging functions for GitHub Actions workflows
+
+# Output an informational message to stderr
+log_info() {
+  local message="$1"
+  echo "$message" >&2
+}
```
fantapop: It's surprising to me that all these need to go to stderr.  Is it because some script that's using it, needs stdout for communicating between processes?

fantapop (approved):
This is looking good to me.  It seems like we could get away with logging to stdout for info lines and stderr for warn and error lines.

PR #104, opened 2026-04-10
[CNSL-2134] Add pending deploy branch management to release workflow

linhcrl:
After a successful release (tag creation), the workflow now automatically manages pending deploy branches:

- Finds the most recent merged PR with an automation/pending-deploy-YYYYMMDD-hhmmss head branch to identify what the last pending deploy branch was
- Compares its timestamp with any current live pending deploy branch
- Creates a new pending deploy branch if a newer one was already merged to main (indicating the current live branch is stale), or if none exists
- Retargets all open PRs from old pending deploy branches to the current one. (As of now, workflow does not resolve conflicts caused by this.)

This ensures that after each release, there's always a fresh pending deploy branch based on the latest main, and PRs are automatically pointed to the right target.

Example job run when no open PRs: https://github.com/cockroachlabs/cockroach-cloud-sdk-go-automation-testing/actions/runs/25906490925/job/76141363111

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,13 +4,15 @@ on:
   push:
     branches: [main]
 
-# contents:write is required for autotag-from-changelog to push tags.
 permissions:
-  contents: write
+  contents: write  # Push tags and create branches
+  pull-requests: write  # Retarget PRs to pending deploy branch
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
+      tag-created: ${{ steps.autotag.outputs.tag-created }}
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -35,3 +39,150 @@ jobs:
           gh api repos/cockroachdb/ccloud-private/dispatches \
             -f event_type=sdk-release \
             -f "client_payload[version]=${{ steps.autotag.outputs.tag }}"
+
+  manage-pending-deploy:
```
fantapop, line 50: Can you comment a high level overview of how this works here?  The Step comments are helpful but they feel piecemeal and it's hard to get an overall understanding of what we're doing here by following them.

fantapop (approved):
This all seems pretty reasonable to me.  I don't have a good feeling on whether it will work. It's a lot of bash inlined in strings to be looking at.  I think it would benefit from pulling out into functions and having tests for it but given that our timeline is pretty short, I'm okay pushing that off for later in case we have trouble maintaining it.    

I did add one comment about documenting better overall what the workflow does at the top level.

.github/workflows/release.yml
linhcrl: done!

PR #105, opened 2026-04-11
[CNSL-1934] Add automated pending deploy branch management

linhcrl:
Introduces two workflows to manage the release process:

1. pending-deploy-pr.yml (workflow_dispatch trigger):
   - Finds the latest automation/pending-deploy-YYYYMMDD-hhmmss branch
   - Creates a PR to merge it into main
   - Triggers a pending deploy check

2. pending-deploy-check.yml (pull_request and workflow_dispatch trigger):
   - Validates pending deploy PRs before merge
   - Checks that Managed-service-commit-SHA trailers reference deployed commits
   - Blocks merge until all changes are confirmed deployed in managed-service
   - Posts PR comments detailing any undeployed commits

This ensures SDK releases only reference deployed CC API changes.

------

## Testing

Did a mini bug bash testing the SDK release workflows end to end. Here are some test cases I tested for this workflow: https://docs.google.com/spreadsheets/d/1_SJbN6T9XU3_r01guk-KrCYoLQocs6ECaRNghr28X5A?gid=1453507795#gid=1453507795

Example PR: https://github.com/cockroachlabs/cockroach-cloud-sdk-go-automation-testing/pull/184

------

#### PR opened by pending-deploy-pr.yml (before all the problematic commits got removed)
Screenshot 2026-05-15 at 2 58 59 AM

#### Comment left by pending-deploy-check.yml on failed check (before the problematic commits got removed)
Screenshot 2026-05-15 at 2 59 10 AM

### DIFF
--- a/.github/workflows/pending-deploy-check.yml
+++ b/.github/workflows/pending-deploy-check.yml
@@ -1,0 +1,92 @@
+name: Pending Deploy Check
+
+# This workflow validates that pending deploy PRs are safe to merge.
+#
+# Trigger: Pull requests to main from automation/pending-deploy-* branches
+# Purpose: Verify all commits in the PR have been deployed in managed-service
+#
+# Why: SDK changes are generated from managed-service OpenAPI specs. We must ensure
+# the corresponding managed-service changes have been deployed before merging the SDK changes,
+# otherwise the SDK could reference unreleased API features.
+#
+# Flow:
+# 1. For each commit in the PR, extract the Managed-service-commit-SHA trailer
+# 2. Check if that SHA is in the latest managed-service release tag
+# 3. If any commits are not yet deployed, post a detailed comment and fail the PR
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Pending deploy branch to check'
+        required: true
+        type: string
+      pr_url:
+        description: 'PR URL to comment on if check fails'
+        required: true
+        type: string
+
+permissions:
+  contents: read  # Checkout repository and read commit history
+  pull-requests: write  # Post failure comments on PRs
+
+jobs:
+  pending-deploy-check:
+    runs-on: ubuntu-latest
+    if: |
+      (github.event_name == 'pull_request' && startsWith(github.head_ref, 'automation/pending-deploy-')) ||
+      (github.event_name == 'workflow_dispatch' && startsWith(inputs.branch, 'automation/pending-deploy-'))
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Validate workflow input
+        if: github.event_name == 'workflow_dispatch'
+        id: validate-input
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          source scripts/lib/actions-helpers.sh
+
+          # Extract PR number from URL
+          PR_NUMBER=$(echo "${{ inputs.pr_url }}" | grep -oE '[0-9]+$')
+          set_output pr_number "$PR_NUMBER"
+
+          # Validate the PR's head branch matches the input branch
+          PR_HEAD_BRANCH=$(gh pr view "$PR_NUMBER" --json headRefName --jq '.headRefName')
+          if [ "$PR_HEAD_BRANCH" != "${{ inputs.branch }}" ]; then
+            log_error "PR #$PR_NUMBER head branch ($PR_HEAD_BRANCH) does not match input branch (${{ inputs.branch }})"
+            exit 1
+          fi
+
+      - name: Run deployment check
+        id: check-deployment
+        env:
+          PENDING_DEPLOY_BRANCH: ${{ inputs.branch || github.head_ref }}
+          MANAGED_SERVICE_TOKEN: ${{ secrets.MANAGED_SERVICE_TOKEN }}
+        run: scripts/pending-deploy-check.sh
+
+      - name: Post failure comment and fail
+        if: steps.check-deployment.outputs.has_issues == 'true'
+        env:
+          PR_NUMBER: ${{ steps.validate-input.outputs.pr_number || github.event.pull_request.number }}
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          scripts/post-failure-comment.sh
+          source scripts/lib/actions-helpers.sh
+          log_error "Deployment check failed - see PR comment for details"
+          exit 1
+
+      - name: Summary
+        if: always()
+        run: |
+          source scripts/lib/actions-helpers.sh
+          if [ "${{ steps.check-deployment.outputs.has_issues }}" != "true" ]; then
+            log_info "All commits in pending deploy branch are deployed in managed-service"
+          else
+            log_info "Some commits are not deployed or potentially not deployed"
+          fi
--- a/.github/workflows/pending-deploy-pr.yml
+++ b/.github/workflows/pending-deploy-pr.yml
@@ -1,0 +1,83 @@
+name: Pending Deploy PR
+
+# This workflow creates PRs to apply pending deploy changes to main.
+#
+# Trigger: workflow_dispatch - called by TeamCity when managed-service is deployed
+# Purpose: Automatically create a PR to merge the latest automation/pending-deploy-* branch into main
+#
+# Flow:
+# 1. Find the latest automation/pending-deploy-YYYYMMDD-HHMMSS branch
+# 2. Check if it has commits not in main
+# 3. Create a PR if one doesn't already exist
+
+on:
+  workflow_dispatch:
+    inputs:
+      timestamp:
+        description: 'Deployment timestamp'
+        required: true
+        type: string
+      commit_sha:
+        description: 'Deployed commit SHA'
+        required: true
+        type: string
+
+permissions:
+  contents: read  # Checkout repository and read branches/tags
+  pull-requests: write  # Create and update PRs
+  actions: write  # Trigger pending deploy check workflow
+
+jobs:
+  pending-deploy-pr:
+    runs-on: ubuntu-latest
+    outputs:
+      branch: ${{ steps.create-pr.outputs.branch }}
+      pr_url: ${{ steps.create-pr.outputs.pr_url }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0  # Need full history to compare branches and commits
+
+      - name: Run pending deploy PR workflow
+        id: create-pr
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: scripts/pending-deploy-pr.sh
+
+      # GitHub Actions doesn't automatically trigger workflows when a PR is created using GITHUB_TOKEN
+      # (to prevent recursive workflow triggers). Since we need deployment validation to run,
+      # we explicitly dispatch the check workflow here.
+      - name: Trigger pending deploy check
+        if: steps.create-pr.outputs.pr_url != '' && steps.create-pr.outputs.branch != ''
+        id: trigger-check
+        env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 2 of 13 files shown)


.github/workflows/pending-deploy-check.yml
```diff
@@ -0,0 +1,242 @@
...
+          cat commits.txt
+
+      - name: Get latest managed-service release tag
+        id: get-release-tag
+        if: steps.check-branch.outputs.is_pending_deploy == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.MANAGED_SERVICE_TOKEN }}
+        run: |
+          # Get all release tags matching release-YYYY-MM-DD-N pattern
+          all_tags=$(gh api repos/cockroachlabs/managed-service/tags --paginate --jq '.[].name' | grep -E '^release-[0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]+$')
```
fantapop: what does --paginate do here?  Does that restrict us to the first page?

.github/workflows/pending-deploy-check.yml
```diff
@@ -0,0 +1,242 @@
...
+        if: steps.check-branch.outputs.is_pending_deploy == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.MANAGED_SERVICE_TOKEN }}
+        run: |
+          latest_tag="${{ steps.get-release-tag.outputs.tag }}"
+
+          # Initialize result files
+          > not_deployed.txt
+          > missing_trailer.txt
+          > unexpected_status.txt
```
fantapop: This is kind of a weird syntax... I guess if it works.  Normally I would use  touch <filename>

.github/workflows/pending-deploy-check.yml
```diff
@@ -0,0 +1,242 @@
...
+
+          # Process each commit
+          while IFS='|' read -r sha subject; do
+            echo "Checking commit $sha: $subject"
+
+            # Extract Managed-service-commit-SHA trailer from commit message
```
fantapop: I'm not really clear whats going on here.  Is trailer the right terminology to be used?

.github/workflows/pending-deploy-pr.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0  # Need full history to compare branches and commits
+
+      - name: Fetch latest remote refs
+        run: |
+          # Ensure we have the latest branches and tags from origin
+          git fetch origin --tags --force
+          git fetch origin --prune
```
fantapop: I was surprised this step would be necessary in either workflow.  Can you document why that is in a comment?

.github/workflows/pending-deploy-pr.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Find latest pending deploy branch
```
fantapop: It looks like a lot of the stuff from this workflow are copied from the other workflow.  Maybe it would make sense to bundle this stuff into an action instead.

fantapop (commented):
I left some feedback.  In general, this feels like a lot of logic to be inlining into these workflows and going untested.  Did you consider the approach we took in the actions repo to separate out some of the logic to actual code files?

.github/workflows/pending-deploy-check.yml
linhcrl: added a comment to explain why it's necessary

.github/workflows/pending-deploy-check.yml
linhcrl: updated to touch

.github/workflows/pending-deploy-pr.yml
linhcrl: removed

.github/workflows/pending-deploy-pr.yml
linhcrl: It doesn't seem like there's enough overlap to warrant shared actions but I moved things into shell scripts for better readability and left comments to explain what's happening

.github/workflows/pending-deploy-check.yml
linhcrl: "Trailer" should be the correct Git terminology here. Git commit trailers are standardized key-value pairs at the end of commit messages (like Signed-off-by: or Co-Authored-By:).    

 See the [git-interpret-trailers](https://git-scm.com/docs/git-interpret-trailers) for more info.
 
----

**Context:**
SDK commits include a Managed-service-commit-SHA: <sha> trailer that links back to the managed-service commit that generated the SDK changes. 

**What this part of the code does:**
For each commit in the pending deploy branch, this section:                                                                                                                                                                                                                                                    
  1. Extracts the managed service commit SHA trailer      
  2. If the trailer is missing, the commit gets categorized in `missing_trailer.txt` (these are commits we can't verify the deployment status of)                     
  3. If the trailer exists, it checks whether that managed-service SHA has been deployed by verifying it's in the latest release tag. If the SHA has not been deployed yet, we write it to the `not_deployed.txt` file. If we can't determine deployment status, we write it to `unexpected_status.txt`. 
  4. At the end, we look at the contents of `missing_trailer.txt`, `not_deployed.txt`, and `unexpected_status.txt`. If any of them contain data, we fail the PR and comment on the PR with information about why the check failed. I included a screenshot of what this looks like

linhcrl: #### Comment left by pending-deploy-check.yml on failed check
Screenshot 2026-04-28 at 9 10 05 PM

scripts/pending-deploy-pr.sh
```diff
@@ -0,0 +1,159 @@
...
+  log_info "Created PR: $pr_url"
+  echo "pr_url=$pr_url" >> "$GITHUB_OUTPUT"
+}
+
+# Main execution
+main() {
+  log_info "=== Starting pending deploy PR workflow ==="
+
+  if [[ -z "${GITHUB_TOKEN:-}" ]]; then
+    log_error "GITHUB_TOKEN environment variable is not set"
```
fantapop: nit: how about:
```suggestion
    log_error "GITHUB_TOKEN environment variable is required to be set"
```

etc...

scripts/pending-deploy-pr.sh
```diff
@@ -0,0 +1,159 @@
...
+
+  log_info "Checking for commits in $branch not in main"
+
+  # Count commits in branch but not in origin/main
+  local commit_count
+  commit_count=$(git rev-list --count origin/main..origin/"$branch")
+
+  if [[ "$commit_count" -eq 0 ]]; then
+    log_info "No new commits in $branch"
+    echo "has_commits=false" >> "$GITHUB_OUTPUT"
```
fantapop: nit: how about `has_new_commits`?

scripts/pending-deploy-pr.sh
```diff
@@ -0,0 +1,159 @@
...
+    --title "Applying pending deploy changes: $branch" \
+    --body "$pr_body")
+
+  if [[ -z "$pr_url" ]]; then
+    log_error "Failed to create PR"
+    exit 1
+  fi
+
+  log_info "Created PR: $pr_url"
+  echo "pr_url=$pr_url" >> "$GITHUB_OUTPUT"
```
fantapop: nit: since you're already including logging helpers everywhere it would be nice to use one for these as well. I think we have `set_output` in the actions repo.

fantapop (commented):
I left a few nits on this.  It would be nice to generate some tests as well for what we can.

scripts/pending-deploy-pr.sh
linhcrl: updated

scripts/pending-deploy-pr.sh
linhcrl: sounds good 👍

scripts/pending-deploy-pr.sh
linhcrl: updated

linhcrl:
> I left a few nits on this. It would be nice to generate some tests as well for what we can.

added some tests

scripts/lib/pending-deploy-helpers-test.sh
```diff
@@ -0,0 +1,374 @@
...
+      else
+        command git "$@"
+      fi
+    }
+    export -f git
+
+    check_all_commits "release-2024-01-01-1"
+
+    # Check that commit was added to missing_trailer.txt
+    grep -q "abc123|Add new feature" missing_trailer.txt
```
fantapop: is there a long opt we can use?

scripts/lib/pending-deploy-helpers.sh
```diff
@@ -0,0 +1,261 @@
...
+    exit 1
+  fi
+
+  log_info "Latest pending deploy branch: $latest"
+  set_output branch "$latest"
+  export BRANCH="$latest"
+}
+
+# Check if the branch has commits not in main
+check_commits() {
```
fantapop: nit: This name seems too similar to the `check_all_commits` function above.  Are there some more specific names we can use?

scripts/lib/pending-deploy-helpers.sh
```diff
@@ -0,0 +1,261 @@
...
+# Create or check for existing PR
+create_pr_if_not_exists() {
+  local branch="$1"
+  local has_new_commits="$2"
+
+  if [[ -z "$branch" ]] || [[ "$has_new_commits" != "true" ]]; then
+    return 0
+  fi
+
+  export GH_TOKEN="$GITHUB_TOKEN"
```
fantapop: nit: If possible we should do this further up in the call chain.  Having such similar env vars we're shuffling between down here is not ideal.

scripts/lib/validation.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# Validation helper functions
+
+# Check if required commands are available in PATH
+# Usage: check_required_commands "cmd1" "cmd2" "cmd3"
+check_required_commands() {
```
fantapop, line 6: I see this in the #108 pr as well.   Looks like you'll have to do some rebasing.

scripts/pending-deploy-check.sh
```diff
@@ -0,0 +1,73 @@
...
+    log_error "GITHUB_OUTPUT environment variable is required"
+    exit 1
+  fi
+
+  # Get list of commits in pending deploy branch not yet in main
+  get_commits "$PENDING_DEPLOY_BRANCH"
+
+  # Fetch the latest managed-service release tag for comparison
+  if ! get_latest_release_tag; then
+    exit 1
```
fantapop, line 42: Should we log an error here?

scripts/pending-deploy-pr.sh
```diff
@@ -0,0 +1,56 @@
...
+
+  if [[ -z "${GITHUB_REPOSITORY:-}" ]]; then
+    log_error "GITHUB_REPOSITORY environment variable is required"
+    exit 1
+  fi
+
+  if [[ -z "${GITHUB_OUTPUT:-}" ]]; then
+    log_error "GITHUB_OUTPUT environment variable is required"
+    exit 1
+  fi
```
fantapop: nit: I think I've seen this pattern a few times.  We could have a check_required_env vars similar to check_required_params.

fantapop (approved):
I've left some nits and small feedback.  signing off though.

scripts/pending-deploy-pr.sh
linhcrl: Added `check_required_env` and added some test cases! Great idea, thanks!

scripts/pending-deploy-check.sh
linhcrl: done!

scripts/lib/pending-deploy-helpers.sh
linhcrl: moved it to right after the validation step

scripts/lib/pending-deploy-helpers.sh
linhcrl: renamed to `check_branch_has_new_commits`

scripts/lib/pending-deploy-helpers-test.sh
linhcrl: updated

PR #108, opened 2026-04-25
[CNSL-1930] Add automated OpenAPI spec sync workflow

linhcrl:
This workflow automates synchronization of OpenAPI specs from the managed-service repository. When triggered via workflow_dispatch with `openapi-spec-changed` or `openapi-spec-merged` event types:

- Fetches the latest OpenAPI spec from the managed-service PR branch (for changed events) or from the specific commit SHA (for merged events)
- Regenerates SDK client code using make generate-openapi-client
- Invokes Claude AI (via Vertex AI) to analyze changes and update CHANGELOG.md with appropriate entries
- Creates or updates PR from bot fork to pending-deploy branch

The workflow includes scripts for finding corresponding SDK PRs and invoking Claude for changelog generation. The Claude prompt instructs it to analyze git diffs, update CHANGELOG.md following Keep a Changelog conventions, identify breaking changes, and output structured metadata for commit messages and PR descriptions.

----

## Testing

Did a mini bug bash testing the SDK release workflows end to end. Here are some test cases I tested for this workflow: https://docs.google.com/spreadsheets/d/1_SJbN6T9XU3_r01guk-KrCYoLQocs6ECaRNghr28X5At?gid=0#gid=0

Example PR opened by workflow 
- On `openapi-spec-changed` event: https://github.com/cockroachlabs/cockroach-cloud-sdk-go-automation-testing/pull/178
- On `openapi-spec-merged` event: https://github.com/cockroachlabs/cockroach-cloud-sdk-go-automation-testing/pull/179

Note: in the case of the _"merged"_ event, there's an additional trailer in the commit message

### DIFF
--- a/.claude/skills/generate-sync-metadata.md
+++ b/.claude/skills/generate-sync-metadata.md
@@ -1,0 +1,70 @@
+# Generate Sync Metadata
+
+This skill is ONLY for the automated OpenAPI sync workflow (scripts/openapi-sync.sh).
+Do not use this skill for manual changelog updates or other changelog work.
+
+Analyze the OpenAPI spec sync changes, update CHANGELOG.md, and generate structured metadata (commit title, commit body, PR title, PR description) for the sync.
+
+## CRITICAL: Source of Truth
+
+**The git diff output is your ONLY source of truth.**
+- Document ONLY what appears in the diff - nothing more, nothing less
+- If something is not in the diff, do NOT document it
+- All changelog entries, commit messages, and PR descriptions must be based ONLY on what you see in the diff
+
+## Instructions
+
+**Step 1: View the changes**
+- Run `printenv BASE_BRANCH` and save the exact output value
+- Run `git diff origin/<BASE_BRANCH_VALUE>` where you substitute `<BASE_BRANCH_VALUE>` with the EXACT value from the printenv command
+  - Example: if printenv output "automation/pending-deploy-20260514-043826", run `git diff origin/automation/pending-deploy-20260514-043826`
+  - Example: if printenv output "main", run `git diff origin/main`
+- **DO NOT** run `git diff main` or `git diff origin/main` IF BASE_BRANCH is NOT `main`
+- **DO NOT** run `git log`, `git show`, or any other git command
+
+**Step 2: Update CHANGELOG.md**
+- Follow the changelog conventions documented in CLAUDE.md
+- IMPORTANT: Only modify CHANGELOG.md. Do not modify any other files including generated code, spec files, or configuration files
+
+**Step 3: Generate commit and PR metadata**
+- Follow the guidelines in the "Output Format" section below
+
+## Output Format
+
+End your response with these exact markers and formatting:
+
+```
+---COMMIT_TITLE---
+<one line, max 72 characters>
+---COMMIT_BODY---
+<multi-line, wrap at 72 characters>
+---PR_TITLE---
+<one line, max 72 characters>
+---PR_DESCRIPTION---
+<multi-line, no wrap limit>
+```
+
+Do not add any text after the PR description section.
+
+## Writing Guidelines
+
+**For Commit Body:**
+- Describe the actual SDK/API changes from a user perspective
+- Focus on WHAT changed in the API/SDK, not the process of updating files
+- DO NOT include file lists - focus on user-facing API/SDK changes
+- Provide a high-level summary (e.g., what new operations are available, what fields were added/modified)
+- Group related changes together (e.g., "Added JWT issuer management operations including create, get, list, update, and delete")
+- For breaking changes, clearly call out what changed and why it might break existing code
+
+**For PR Description:**
+- Start with a reader-friendly summary (2-3 sentences)
+- Describe the actual SDK/API changes from a user perspective, similar to the commit body
+- DO NOT include file lists - focus on user-facing API/SDK changes
+- Can be more detailed than commit body - include context that helps reviewers understand the changes
+
+**For Commit Titles and PR Titles:**
+- Use imperative mood (e.g., "Add", "Update", "Fix", not "Added", "Updated", "Fixed")
+- DO NOT write titles like "Update CHANGELOG.md" - the changelog update is just a side effect, not the main change
+- Titles must be under 72 characters (strict limit)
+- If changes are focused on one specific area, describe that clearly (e.g., "Add MFA audit log actions", "Update invoice field descriptions")
+- For multiple unrelated changes: review your title against all affected endpoints and fields. If your title doesn't cover everything or exceeds 72 characters, use the generic title "Sync OpenAPI spec from managed-service" and describe the changes in the commit body instead.
--- a/.github/workflows/openapi-sync.yml
+++ b/.github/workflows/openapi-sync.yml
@@ -1,0 +1,97 @@
+name: Sync OpenAPI Spec from Managed Service
+
+on:
+  workflow_dispatch:
+    inputs:
+      event_type:
+        description: 'Event type (openapi-spec-changed or openapi-spec-merged)'
+        required: true
+        type: choice
+        options:
+          - openapi-spec-changed
+          - openapi-spec-merged
+      pr_url:
+        description: 'Managed-service PR URL'
+        required: true
+        type: string
+      sha:
+        description: 'Managed-service commit SHA (required for merged events)'
+        required: false
+        type: string
+
+permissions:
+  contents: write       # Push commits to fork branch
+  pull-requests: write  # Create/update pull requests in upstream repo
+  id-token: write       # Authenticate to Google Cloud for Vertex AI
+
+jobs:
+  openapi-sync:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_REPOSITORY: ${{ github.repository }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Validate branch
+        run: |
+          source scripts/lib/actions-helpers.sh
+          if [[ "${{ github.ref }}" != "refs/heads/main" ]]; then
+            log_error "This workflow can only be triggered from the main branch (current ref: ${{ github.ref }})"
+            exit 1
+          fi
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod  # Required by make generate-openapi-client and make validate
+
+      - name: Authenticate to Google Cloud (Vertex AI)
+        uses: google-github-actions/auth@v3
+        with:
+          project_id: vertex-model-runners
+          service_account: vertex-ai-user@vertex-model-runners.iam.gserviceaccount.com
+          workload_identity_provider: projects/108106229451/locations/global/workloadIdentityPools/vertex-ai-user/providers/vertex-ai-user
+
+      - name: Generate updated client from OpenAPI spec
+        id: sync
+        env:
+          EVENT_TYPE: ${{ github.event.inputs.event_type }}
+          MANAGED_SERVICE_PR_URL: ${{ github.event.inputs.pr_url }}
+          MANAGED_SERVICE_SHA: ${{ github.event.inputs.sha }}
+          MANAGED_SERVICE_TOKEN: ${{ secrets.MANAGED_SERVICE_TOKEN }}
+          FORK_OWNER: crl-gh-actions-pr-bot
+          FORK_PUSH_TOKEN: ${{ secrets.FORK_PUSH_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
+        run: scripts/openapi-sync-generate.sh
+
+      - name: Update changelog with Claude
+        id: changelog
+        if: steps.sync.outputs.has_changes == 'true'
+        env:
+          BASE_BRANCH: ${{ steps.sync.outputs.base_branch }}
+          HAS_CHANGES: ${{ steps.sync.outputs.has_changes }}
+        run: scripts/openapi-sync-update-changelog.sh
+
+      - name: Publish changes (commit and create PR)
+        if: steps.sync.outputs.has_changes == 'true' || github.event.inputs.event_type == 'openapi-spec-merged'
+        env:
+          EVENT_TYPE: ${{ github.event.inputs.event_type }}
+          MANAGED_SERVICE_PR_URL: ${{ github.event.inputs.pr_url }}
+          MANAGED_SERVICE_SHA: ${{ github.event.inputs.sha }}
+          FORK_OWNER: crl-gh-actions-pr-bot
+          FORK_PUSH_TOKEN: ${{ secrets.FORK_PUSH_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
+          BASE_BRANCH: ${{ steps.sync.outputs.base_branch }}
+          HAS_CHANGES: ${{ steps.sync.outputs.has_changes }}
+          EXISTING_SDK_PR: ${{ steps.sync.outputs.existing_sdk_pr }}
+          HEAD_BRANCH: ${{ steps.sync.outputs.head_branch }}
+          SDK_PR_NUMBER: ${{ steps.sync.outputs.sdk_pr_number }}
+          COMMIT_TITLE: ${{ steps.changelog.outputs.commit_title }}
+          COMMIT_BODY: ${{ steps.changelog.outputs.commit_body }}
+          PR_TITLE: ${{ steps.changelog.outputs.pr_title }}
+          PR_DESCRIPTION: ${{ steps.changelog.outputs.pr_description }}
+        run: scripts/openapi-sync-publish.sh
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3,3 +3,38 @@
 ## GitHub Actions
 
 - When writing or modifying GitHub Actions workflows, always look up the latest major version of each action before using it. Do not assume you know the current version.
+
+## Changelog
+
+Follow Keep a Changelog conventions when updating CHANGELOG.md.
+
+**Unreleased Section Structure:**
+- The `## [Unreleased]` section starts with the `## [Unreleased]` header and ends when the next `##` level header appears (typically a version like `## [x.y.z]`)
+- It contains subsections: `### Added`, `### Changed`, `### Deprecated`, `### Removed`, `### Fixed`, `### Security`
+- Subsections must appear in this exact order (top to bottom): Added, Changed, Deprecated, Removed, Fixed, Security
+- There should be only ONE of each subsection type within `## [Unreleased]`
+
+**Adding Entries:**
+- Add all new entries under the `## [Unreleased]` section
+- Choose the appropriate subsection based on the change type:
+  - Added: new features
+  - Changed: changes in existing functionality
+  - Deprecated: soon-to-be removed features
+  - Removed: now removed features
+  - Fixed: bug fixes
+  - Security: vulnerabilities
+- If the subsection already exists: add your entry at the top of that subsection's list - DO NOT create a duplicate subsection header
+- If the subsection doesn't exist: create it within the `## [Unreleased]` section, positioned according to the order above (Added, Changed, Deprecated, Removed, Fixed, Security)
+
+**How to Write Changelog Entries:**
+- Be brief. One sentence per entry.
+- Write from the user's perspective. Describe what they can now do, not what code changed.
+- Start each entry with a verb (e.g. Add, Update, Fix, Remove).
+- Prefix breaking changes with "Breaking Change: ".
+- Group related changes into one entry when they form a single feature (e.g. "Add support for folder management operations").
+- Don't list individual fields, models, or implementation details unless they're the primary change.
+- For OpenAPI spec syncs: mention the new operations and capabilities, not the sync process itself.
+
+**Examples:**
+- Good: `Add support for JWT issuer management operations (CreateJwtIssuer, GetJwtIssuer, ListJwtIssuers, UpdateJwtIssuer, DeleteJwtIssuer).`
+- Bad: `Sync OpenAPI spec from managed-service PR #1234 and add JwtIssuer CRUD operations with Audience, IssuerUrl, JwksUri, and ClaimMap fields.`
--- a/scripts/git-askpass.sh
+++ b/scripts/git-askpass.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+# GIT_ASKPASS script for git authentication. Reads credentials from
+# environment variables so the token is never written to disk.
+set -euo pipefail
+
+# Get the script directory to find helper functions
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source "$SCRIPT_DIR/lib/actions-helpers.sh"
+
+if [ -z "$GIT_USER" ] || [ -z "$GIT_PASSWORD" ]; then
+  log_error "GIT_USER and GIT_PASSWORD are required"
+  exit 1
+fi
+
+case "$1" in
+  Username*) echo "$GIT_USER" ;;
+  Password*) echo "$GIT_PASSWORD" ;;
+esac
--- a/scripts/lib/openapi-sync-helpers.sh
+++ b/scripts/lib/openapi-sync-helpers.sh
@@ -1,0 +1,512 @@
+#!/usr/bin/env bash
+# Helper functions for OpenAPI sync workflow
+
+# Set up git authentication using GIT_ASKPASS
+setup_git_auth() {
+  local user="$1"
+  local password="$2"
+  export GIT_ASKPASS="$SCRIPT_DIR/git-askpass.sh"
+  export GIT_USER="$user"
+  export GIT_PASSWORD="$password"
+  export GIT_TERMINAL_PROMPT=0
+}
+
+# Find an SDK PR that corresponds to a managed-service PR URL.
+#
+# This searches all open PRs in the current repository and inspects
+# all commits in each PR to find one containing a trailer:
+#   Managed-service-pr-url: <pr_url>
+#
+# Args: $1 - managed-service PR URL
+# Exports on success:
+#   SDK_PR_NUMBER: The SDK PR number
+#   HEAD_BRANCH: The SDK PR head branch name
+#   BASE_BRANCH: The SDK PR base branch name
+# Returns: 0 if found, 1 if not found
+find_sdk_pr() {
+  local managed_service_pr_url="$1"
+
+  if [[ -z "${managed_service_pr_url:-}" ]]; then
+    log_error "managed_service_pr_url is required"
+    return 1
+  fi
+
+  check_required_env GH_TOKEN GITHUB_REPOSITORY || return 1
+
+  log_info "Searching for SDK PR corresponding to: $managed_service_pr_url"
+
+  # Get all open SDK PRs
+  local sdk_open_prs
+  sdk_open_prs=$(gh pr list --repo "$GITHUB_REPOSITORY" --state open --json number --jq '.[].number')
+
+  for sdk_pr_num in $sdk_open_prs; do
+    log_info "Checking SDK PR #$sdk_pr_num"
+
+    # Get all commits in this SDK PR
+    local sdk_commits
+    sdk_commits=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$sdk_pr_num/commits" --jq '.[].sha')
+
+    # Check each commit's message for the trailer
+    for sdk_commit_sha in $sdk_commits; do
+      local sdk_commit_msg
+      sdk_commit_msg=$(gh api "repos/$GITHUB_REPOSITORY/git/commits/$sdk_commit_sha" --jq '.message')
+
+      if echo "$sdk_commit_msg" | grep --quiet --fixed-strings "Managed-service-pr-url: $managed_service_pr_url"; then
+        log_info "Found matching SDK PR #$sdk_pr_num (commit $sdk_commit_sha)"
+
+        # Get the SDK PR details
+        local pr_data
+        pr_data=$(gh pr view "$sdk_pr_num" --repo "$GITHUB_REPOSITORY" --json headRefName,baseRefName)
+
+        # Export results
+        export SDK_PR_NUMBER="$sdk_pr_num"
+        export HEAD_BRANCH=$(echo "$pr_data" | jq --raw-output '.headRefName')
+        export BASE_BRANCH=$(echo "$pr_data" | jq --raw-output '.baseRefName')
+        return 0
+      fi
+    done
+  done
+
+  log_info "No matching SDK PR found"
+  return 1
+}
+
+# Validate event payload business rules
+validate_event_payload() {
+  # Validate event_type is one of the allowed values
+  if [[ "$EVENT_TYPE" != "openapi-spec-changed" && "$EVENT_TYPE" != "openapi-spec-merged" ]]; then
+    log_error "Invalid 'event_type': $EVENT_TYPE. Must be 'openapi-spec-changed' or 'openapi-spec-merged'"
+    exit 1
+  fi
+
+  if [[ "$EVENT_TYPE" == "openapi-spec-merged" && -z "${MANAGED_SERVICE_SHA:-}" ]]; then
+    log_error "Input 'sha' is required for openapi-spec-merged event"
+    exit 1
+  fi
+
+  if [[ "$EVENT_TYPE" == "openapi-spec-changed" && -n "${MANAGED_SERVICE_SHA:-}" ]]; then
+    log_error "Input 'sha' should not be provided for openapi-spec-changed event"
+    exit 1
+  fi
+}
+
+# Parse managed-service PR metadata from URL
+#
+# Validates that the URL is from cockroachlabs/managed-service and extracts
+# the PR number.
+#
+# Exports:
+#   MS_OWNER: The GitHub organization
+#   MS_REPO: The repository name
+#   MS_PR_NUMBER: The managed-service PR number
+parse_managed_service_pr_url() {
+  MS_OWNER="cockroachlabs"
+  MS_REPO="managed-service"
+
+  # Validate URL format and extract PR number
+  if [[ ! "$MANAGED_SERVICE_PR_URL" =~ github\.com/$MS_OWNER/$MS_REPO/pull/([0-9]+) ]]; then
+    log_error "Invalid 'pr_url'. Must be a valid GitHub PR URL from $MS_OWNER/$MS_REPO"
+    exit 1
+  fi
+
+  MS_PR_NUMBER="${BASH_REMATCH[1]}"
+
+  log_info "Managed-service PR: $MS_OWNER/$MS_REPO#$MS_PR_NUMBER"
+
+  # Export for use in other functions
+  export MS_OWNER MS_REPO MS_PR_NUMBER
+}
+
+# Configure Git
+configure_git() {
+  git config user.name "$FORK_OWNER"
+  git config user.email "$FORK_OWNER@users.noreply.github.com"
+
+  FORK_URL="https://github.com/$FORK_OWNER/$(basename "$GITHUB_REPOSITORY").git"
+  log_info "Adding fork remote: $FORK_URL"
+  git remote add fork "$FORK_URL"
+
+  # Get the default branch name
+  local default_branch
+  default_branch=$(gh repo view "$GITHUB_REPOSITORY" --json defaultBranchRef --jq '.defaultBranchRef.name')
+  log_info "Default branch: $default_branch"
+
+  # Sync fork's default branch with upstream to include any changes from upstream
+  # that the fork doesn't have. This prevents feature branches from appearing to add
+  # files requiring special permissions (e.g., workflows), which would cause permission issues
+  log_info "Syncing fork's $default_branch with origin/$default_branch"
+
+  # Set up askpass for fork push
+  setup_git_auth "$FORK_OWNER" "$FORK_PUSH_TOKEN"
+
+  git push --force fork "origin/$default_branch:$default_branch"
+}
+
+# Determine head and base branches for the SDK PR
+# For existing SDK PRs: uses the PR's head branch and base branch
+# For new SDK PRs: finds/creates a base branch, then creates a new head branch from it
+# on openapi-spec-changed: may create a new head branch if no SDK PR exists yet
+# on openapi-spec-merged: the SDK PR must already exist
+#
+# Exports:
+#   HEAD_BRANCH: The SDK PR head branch name
+#   BASE_BRANCH: The SDK PR base branch name (target branch)
+#   EXISTING_SDK_PR: "true" if updating existing PR, "false" if creating new PR
+determine_head_and_base_branches() {
+  # Search all open PRs for one with a commit containing the managed-service PR URL
+  if find_sdk_pr "$MANAGED_SERVICE_PR_URL"; then
+    log_info "Using existing SDK PR branch: $HEAD_BRANCH (targets $BASE_BRANCH)"
+
+    # Set up askpass for fork operations
+    setup_git_auth "$FORK_OWNER" "$FORK_PUSH_TOKEN"
+
+    if ! git fetch fork "$HEAD_BRANCH" 2>&1; then
+      log_error "Could not fetch branch $HEAD_BRANCH from fork"
+      exit 1
+    fi
+    git checkout -B "$HEAD_BRANCH" "fork/$HEAD_BRANCH"
+
+    # Fetch the base branch from origin so Claude can diff against the latest state
+    setup_git_auth "x-access-token" "$GH_TOKEN"
+    git fetch origin "$BASE_BRANCH"
+
+    EXISTING_SDK_PR="true"
+  else
+    # No existing SDK PR found
+    if [[ "$EVENT_TYPE" == "openapi-spec-merged" ]]; then
+      # For merged events, we expect an SDK PR to already exist from the earlier
+      # openapi-spec-changed event. If not found, something went wrong.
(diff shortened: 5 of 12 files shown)


.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,414 @@
...
+
+jobs:
+  sync-openapi:
+    runs-on: ubuntu-latest
+    env:
+      EVENT_TYPE: ${{ github.event.action || github.event.inputs.event_type }}
+      MANAGED_SERVICE_PR_URL: ${{ github.event.client_payload.pr_url || github.event.inputs.pr_url }}
+      MANAGED_SERVICE_SHA: ${{ github.event.client_payload.sha || github.event.inputs.sha }}
+    steps:
+      # Step 1: Validate event payload
```
fantapop: nit: claude is always trying to add step numbers to my comment but it means if we add or remove a step we need to update all the step numbers.  I usually ask for the step numbers to be removed.

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,414 @@
...
+          fi
+
+          echo "Event validation passed"
+          echo "Event type: $EVENT_TYPE"
+          echo "PR URL: $MANAGED_SERVICE_PR_URL"
+          echo "SHA: $MANAGED_SERVICE_SHA"
+
+      # Step 2: Resolve managed-service PR metadata from pr_url
+      - name: Resolve managed-service PR metadata
+        id: ms-pr
```
fantapop: nit: how about just pr?

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,414 @@
...
+            exit 1
+          fi
+
+          MS_OWNER="${BASH_REMATCH[1]}"
+          MS_REPO="${BASH_REMATCH[2]}"
+          MS_PR_NUMBER="${BASH_REMATCH[3]}"
+
+          echo "Managed-service PR: $MS_OWNER/$MS_REPO#$MS_PR_NUMBER"
+
+          echo "Managed-service PR: $MS_OWNER/$MS_REPO#$MS_PR_NUMBER"
```
fantapop: Is this a mistake?  These appear duplicated

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,414 @@
...
+          persist-credentials: false
+
+      # Step 4: Find corresponding SDK PR by searching for trailer
+      - name: Find corresponding SDK PR
+        id: find-sdk-pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set +e  # Don't exit on error
+          SCRIPT_OUTPUT=$(.github/scripts/find-sdk-pr.sh "$MANAGED_SERVICE_PR_URL")
```
fantapop: It seems unexpected to put these scripts under the .github directory.  Do we do that elsewhere?

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,410 @@
...
+          SCRIPT_EXIT_CODE=$?
+          set -e
+
+          if [[ $SCRIPT_EXIT_CODE -eq 0 && -n "$SCRIPT_OUTPUT" ]]; then
+            echo "$SCRIPT_OUTPUT" >> $GITHUB_OUTPUT
+            echo "exists=true" >> $GITHUB_OUTPUT
+          else
+            # No SDK PR found
+            if [[ "$EVENT_TYPE" == "openapi-spec-merged" ]]; then
+              echo "Error: No SDK PR found for openapi-spec-merged event"
```
fantapop: should we use the github action error log format here?  Or maybe stderr?

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,410 @@
...
+      # Step 6: Authenticate to Google Cloud (Vertex AI)
+      - name: Authenticate to Google Cloud (Vertex AI)
+        uses: google-github-actions/auth@v3
+        with:
+          project_id: vertex-model-runners
+          service_account: vertex-ai-user@vertex-model-runners.iam.gserviceaccount.com
+          workload_identity_provider: projects/108106229451/locations/global/workloadIdentityPools/vertex-ai-user/providers/vertex-ai-user
+
+      # Step 7: Set up docker-compose compatibility
+      - name: Set up docker-compose compatibility
```
fantapop: can we instead update the sdk to use "docker compose" instead of adding this?

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,410 @@
...
+          chmod +x "$GIT_ASKPASS"
+
+          if [[ "$EXISTING_SDK_PR_EXISTS" == "true" ]]; then
+            # Use existing PR branch
+            SDK_BRANCH="$EXISTING_SDK_PR_BRANCH"
+            echo "Using existing SDK PR branch: $SDK_BRANCH"
+
+            # Check out the branch from fork
+            git fetch fork "$SDK_BRANCH" || {
+              echo "Error: Could not fetch branch $SDK_BRANCH from fork"
```
fantapop: github actions error format?

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,410 @@
...
+      - name: Set up Claude CLI
+        if: steps.sync.outputs.has_changes == 'true'
+        run: |
+          if command -v roachdev >/dev/null; then
+            printf '#!/bin/sh\nexec roachdev claude -- "$@"\n' > /usr/local/bin/claude
+            chmod +x /usr/local/bin/claude
+            echo "Claude CLI: using roachdev wrapper"
+          else
+            curl --fail --silent --show-error --location https://claude.ai/install.sh | bash -s -- "latest"
+            echo "Claude CLI installed: $(claude --version)"
```
fantapop: since this isn't a reusable action, it seems like you'd either have roachdev or not.  I don't think we should support both.

fantapop (commented):
I left some feedback items here but as I was getting too the bottom, it felt like a lot of logic to be putting in a local workflow file.  Since this is all happening in this repo and we need to check out the repo anyways could this all live in a single build script?   For example take a look at the release script in ccloud cli: https://github.com/cockroachlabs/ccloud-private/blob/main/build/release.sh

.github/workflows/openapi-sync.yml
linhcrl: removed

.github/workflows/openapi-sync.yml
linhcrl: no longer relevant since we don't have steps anymore

.github/workflows/openapi-sync.yml
linhcrl: removed

.github/workflows/openapi-sync.yml
linhcrl: no longer relevant since we reorganized

.github/workflows/openapi-sync.yml
linhcrl: added logging helpers

.github/workflows/openapi-sync.yml
linhcrl: done

.github/workflows/openapi-sync.yml
linhcrl: added logging helper

.github/workflows/openapi-sync.yml
linhcrl: removed the roachdev option

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Authenticate to Google Cloud (Vertex AI)
+        uses: google-github-actions/auth@v3
+        with:
+          project_id: vertex-model-runners
+          service_account: vertex-ai-user@vertex-model-runners.iam.gserviceaccount.com
+          workload_identity_provider: projects/108106229451/locations/global/workloadIdentityPools/vertex-ai-user/providers/vertex-ai-user
+
+      - name: Run OpenAPI sync workflow
+        run: |
+          chmod +x scripts/openapi-sync.sh
```
fantapop: I think we should just add +x to the actual file in the repo so it's checked out with execute

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,333 @@
...
+#   sync-openapi.sh
+#
+# Required environment variables:
+#   EVENT_TYPE: "openapi-spec-changed" or "openapi-spec-merged"
+#   MANAGED_SERVICE_PR_URL: URL of the managed-service PR
+#   MANAGED_SERVICE_SHA: Commit SHA (required for openapi-spec-merged)
+#   MANAGED_SERVICE_TOKEN: GitHub token with managed-service read access
+#   FORK_PUSH_TOKEN: GitHub token with fork push access
+#   FORK_OWNER: GitHub username that owns the fork
+#   GITHUB_TOKEN: GitHub token for creating PRs
```
fantapop: naming seems inconsistent here.  How about CREATE_PR_TOKEN?

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,333 @@
...
+  if [[ -n "${MANAGED_SERVICE_SHA:-}" ]]; then
+    REF="$MANAGED_SERVICE_SHA"
+    log_info "Fetching openapi.json from $MS_OWNER/$MS_REPO at commit $MANAGED_SERVICE_SHA"
+  else
+    REF="refs/pull/$MS_PR_NUMBER/head"
+    log_info "Fetching openapi.json from $MS_OWNER/$MS_REPO PR #$MS_PR_NUMBER"
+  fi
+
+  export GH_TOKEN="$MANAGED_SERVICE_TOKEN"
+  gh api "repos/$MS_OWNER/$MS_REPO/contents/$SPEC_PATH_IN_MS?ref=$REF" \
```
fantapop: We should probably add a check at the top of the file to ensure gh is in the PATH.  I know the actions will have this by default but in case someone tries to run it locally I think it would be good to have it.

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,333 @@
...
+    log_info "No changes to $LOCAL_SPEC_PATH"
+    rm "$LOCAL_SPEC_PATH.new"
+    HAS_CHANGES="false"
+  else
+    log_info "Spec file changed, replacing $LOCAL_SPEC_PATH"
+    mv "$LOCAL_SPEC_PATH.new" "$LOCAL_SPEC_PATH"
+
+    log_info "Running make generate-openapi-client"
+    sudo make generate-openapi-client
+    sudo chown --recursive "$(id --user):$(id --group)" internal/openapi-generator/ pkg/client/ docs/ README.md
```
fantapop: I'm surprised we need sudo here. Is there a way to get around that?

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,333 @@
...
+      log_info "No changes detected, but adding SHA trailer to existing commit"
+
+      CURRENT_COMMIT_MSG=$(git log -1 --format=%B)
+
+      if echo "$CURRENT_COMMIT_MSG" | grep --quiet "Managed-service-commit-SHA: $MANAGED_SERVICE_SHA"; then
+        log_info "SHA trailer already exists in commit"
+      else
+        NEW_COMMIT_MSG=$(printf "%s\nManaged-service-commit-SHA: %s" "$CURRENT_COMMIT_MSG" "$MANAGED_SERVICE_SHA")
+        git commit --amend --message "$NEW_COMMIT_MSG"
+      fi
```
fantapop: can you tell me more about how this happens in the workflow?  I can't quite tell what's going on here.  If a commit has already landed in main we shouldn't try to amend it.

scripts/lib/invoke-claude-changelog.sh
```diff
@@ -0,0 +1,133 @@
...
+#
+# Outputs:
+#   commit_title, commit_body, pr_title, pr_description to stdout
+#   Exit code 0 on success, 1 on failure
+
+MANAGED_SERVICE_PR_URL="$1"
+MANAGED_SERVICE_SHA="${2:-}"
+
+if [[ -z "${MANAGED_SERVICE_PR_URL:-}" ]]; then
+  echo "::error::managed_service_pr_url is required" >&2
```
fantapop: can these use the logging helpers?

scripts/lib/logging.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+# Logging functions for GitHub Actions workflows
+
+# Output an informational message to stderr
+log_info() {
+  local message="$1"
+  echo "$message" >&2
```
fantapop: I think I may have asked this in another PR but it seems like info and notice level logs should go to stdout if possible.

scripts/lib/pending-deploy-helpers.sh
```diff
@@ -0,0 +1,73 @@
...
+#   SDK_PR_BRANCH: The SDK PR branch name
+#
+# Returns:
+#   0 if found, 1 if not found
+
+find_sdk_pr() {
+  local managed_service_pr_url="$1"
+
+  if [[ -z "${managed_service_pr_url:-}" ]]; then
+    log_error "managed_service_pr_url is required"
```
fantapop: do we need to include the logging.sh?  Or is it assumed to included by the invoking script already?  If it's assumed we should probably check that it's available when this script is invoked and fail with a message in case someone tries to use it differently.

scripts/lib/git-askpass.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# GIT_ASKPASS script for fork authentication. Reads credentials from
+# environment variables so the token is never written to disk.
+set -e
+
+if [ -z "$GIT_FORK_USER" ] || [ -z "$GIT_FORK_PASSWORD" ]; then
+  echo "::error::GIT_FORK_USER and GIT_FORK_PASSWORD must be set" >&2
```
fantapop: should this use the logging helpers?

scripts/lib/invoke-claude-changelog.sh
```diff
@@ -0,0 +1,133 @@
...
+PROMPT_TEMPLATE="$SCRIPT_DIR/../prompts/update-changelog.md"
+
+if [[ ! -f "$PROMPT_TEMPLATE" ]]; then
+  echo "::error::Prompt template not found at $PROMPT_TEMPLATE" >&2
+  exit 1
+fi
+
+# Create prompt from template with substitutions
+PROMPT_FILE=$(mktemp)
+sed "s|{{MANAGED_SERVICE_PR_URL}}|$MANAGED_SERVICE_PR_URL|g" "$PROMPT_TEMPLATE" > "$PROMPT_FILE"
```
fantapop: Nit: templatizing with this url doesn't seem to dangerous but in generally it's a pattern that can lead to prompt injection.  Another way you can do this is to make sure the url is in the env that claude is called with and tell it what the env var is in the prompt.   For example

1. Look up the managed service url from the environment using printenv MANAGED_SERVICE_PR_URL
2. ensure.. the trailer exists and contains the text in this exact format but with the value of the env var:  managed-service-url: <MANAGED_SERVICE_PR_URL>
3. etc...

scripts/openapi-sync.sh
linhcrl: The sudo is needed because when GitHub Actions runs this script, it operates as a non-root user. The docker compose commands in the Makefile run containers as root by default, which creates the generated files (in pkg/client/, docs/, etc.) owned by root. The GitHub Actions user then cannot access or commit these files. We use sudo to run the make command, and then sudo chown to change ownership back to the GitHub Actions user so it can stage and commit the changes.

An alternative approach would be to configure the Docker Compose services to run as the current user by setting the user field in docker-compose.yml for both the jq and openapi-generator services. The script would then need to export UID and GID environment variables before calling make generate-openapi-client, and we could remove both sudo calls.

fantapop: okay got it.  Let's not worry about it.  Unless you've already worked on it and have it figured out..

scripts/openapi-sync.sh
linhcrl: renamed

scripts/openapi-sync.sh
linhcrl: I kept sudo for now

scripts/lib/invoke-claude-changelog.sh
linhcrl: updated

scripts/lib/logging.sh
linhcrl: now we are directing info and notice to stdout

scripts/lib/pending-deploy-helpers.sh
linhcrl: i decided to source in each file

scripts/lib/git-askpass.sh
linhcrl: updated

scripts/lib/invoke-claude-changelog.sh
linhcrl: updated

scripts/openapi-sync.sh
linhcrl: I added a comment above the code block to explain what's happening here. Let me know if that helps

scripts/openapi-sync.sh
linhcrl: updated to check using a helper I added called `check_required_commands`

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,60 @@
...
+        required: true
+        type: string
+      sha:
+        description: 'Managed-service commit SHA (required for merged events)'
+        required: false
+        type: string
+
+# Required permissions for the workflow
+permissions:
+  contents: write
```
fantapop: can you add some comments for each one of these that explain why it's needed.

CLAUDE.md
```diff
@@ -3,3 +3,22 @@
...
+- Be brief. One sentence per entry.
+- Write from the user's perspective. Describe what they can now do, not what code changed.
+- Start each entry with a verb (e.g. Add, Update, Fix, Remove).
+- Group related changes into one entry when they form a single feature (e.g. "Add support for folder management operations").
+- Don't list individual fields, models, or implementation details unless they're the primary change.
+- For OpenAPI spec syncs, mention the new operations and capabilities, not the sync process itself. The managed-service PR reference belongs in the commit message,
+not the changelog.
+
+Good: `Add support for JWT issuer management operations (CreateJwtIssuer, GetJwtIssuer, ListJwtIssuers, UpdateJwtIssuer, DeleteJwtIssuer).`
+Bad: `Sync OpenAPI spec from managed-service PR #1234 and add JwtIssuer CRUD operations with Audience, IssuerUrl, JwksUri, and ClaimMap fields.`
```
fantapop: I noticed we have something similar in the skill file as well.  I'm wondering if we can put the instructions for this in it's own skill file and reuse it everywhere.  We'll also need it in ccloud-private.  I don't love the idea of having to check out a shared skills repo everywhere.  But it's probably better than copying the instructions for doing this in multiple places.  We might be able to put it in the actions repo.

scripts/invoke-claude-changelog.sh
```diff
@@ -0,0 +1,137 @@
...
+MANAGED_SERVICE_SHA="${2:-}"
+
+if [[ -z "${MANAGED_SERVICE_PR_URL:-}" ]]; then
+  log_error "managed_service_pr_url is required"
+  exit 1
+fi
+
+log_info "Invoking Claude to analyze changes and update changelog..."
+
+PROMPT_TEMPLATE="$SCRIPT_DIR/../../.claude/skills/update-changelog.md"
```
fantapop: This is a bit surprising.  From the PR it looks like it should be `$SCRIPT_DIR/../.claude/skills...`

scripts/invoke-claude-changelog.sh
```diff
@@ -0,0 +1,137 @@
...
+cp "$PROMPT_TEMPLATE" "$PROMPT_FILE"
+
+# Add SHA trailer instruction if provided
+if [[ -n "${MANAGED_SERVICE_SHA:-}" ]]; then
+  cat >> "$PROMPT_FILE" <<'EOF'
+
+IMPORTANT: The commit body must also include this trailer immediately after the Managed-service-pr-url trailer:
+Managed-service-commit-SHA: $MANAGED_SERVICE_SHA
+EOF
+fi
```
fantapop: It seems like this is no longer necessary?

scripts/invoke-claude-changelog.sh
```diff
@@ -0,0 +1,137 @@
...
+log_info "Claude response received, extracting metadata..."
+
+# Extract commit title (between ---COMMIT_TITLE--- and ---COMMIT_BODY---)
+COMMIT_TITLE=$(echo "$RESULT_TEXT" | awk '/---COMMIT_TITLE---/{flag=1;next}/---COMMIT_BODY---/{flag=0}flag' | grep --invert-match '^[[:space:]]*$' | head --lines=1)
+
+# Extract commit body (between ---COMMIT_BODY--- and ---PR_TITLE---)
+COMMIT_BODY=$(echo "$RESULT_TEXT" | awk '/---COMMIT_BODY---/{flag=1;next}/---PR_TITLE---/{flag=0}flag')
+
+# Extract PR title (between ---PR_TITLE--- and ---PR_DESCRIPTION---)
+PR_TITLE=$(echo "$RESULT_TEXT" | awk '/---PR_TITLE---/{flag=1;next}/---PR_DESCRIPTION---/{flag=0}flag' | grep --invert-match '^[[:space:]]*$' | head --lines=1)
```
fantapop: It's unfortunate that we can't use the autosolve action for this.  A lot of it is repeated logic and somewhat complex as well.  At some point we should talk about what it would take to support this use case.

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,390 @@
...
+  done
+
+  if [[ "$EVENT_TYPE" == "openapi-spec-merged" && -z "${MANAGED_SERVICE_SHA:-}" ]]; then
+    log_error "MANAGED_SERVICE_SHA is required for openapi-spec-merged event"
+    exit 1
+  fi
+}
+
+# Resolve managed-service PR metadata from URL
+resolve_managed_service_pr() {
```
fantapop: how about `parse_managed_service_pr_url` ?  Resolve makes it sound like it's doing some remote fetching of the actual branch

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,390 @@
...
+  log_info "Managed-service PR: $MS_OWNER/$MS_REPO#$MS_PR_NUMBER"
+
+  # Export for use in other functions
+  export MS_OWNER MS_REPO MS_PR_NUMBER
+}
+
+# Configure Git
+configure_git() {
+  git config user.name "crl-console-bot"
+  git config user.email "crl-console-bot@users.noreply.github.com"
```
fantapop: Should this be controlled by the fork user parameter?

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,390 @@
...
+
+  FORK_URL="https://github.com/$FORK_OWNER/$(basename "$GITHUB_REPOSITORY").git"
+  log_info "Adding fork remote: $FORK_URL"
+  git remote add fork "$FORK_URL"
+}
+
+# Determine or create SDK branch
+# Two scenarios:
+# 1. openapi-spec-changed: Create new SDK PR or update existing one
+# 2. openapi-spec-merged: Update existing SDK PR (must already exist)
```
fantapop: The description here confused me a bit.  I think it's just the "Two scenarios" framing.  I misconstrued the 2 scenarios as being creating or updating the branch.  But it's actually that the 2 scenarios are the events.  How about something like this instead? 
```suggestion
# Determine or create SDK branch
# on openapi-spec-changed: the SDK branch may not exist yet. If not, create it. Return the branch.
# on openapi-spec-merged: the SDK branch will already exist so return it.
```

scripts/openapi-sync.sh
```diff
@@ -0,0 +1,390 @@
...
+  # Search all open PRs for one with a commit containing the managed-service PR URL
+  if find_sdk_pr "$MANAGED_SERVICE_PR_URL"; then
+    SDK_BRANCH="$SDK_PR_BRANCH"
+    log_info "Using existing SDK PR branch: $SDK_BRANCH"
+
+    git fetch fork "$SDK_BRANCH" || {
+      log_error "Could not fetch branch $SDK_BRANCH from fork"
+      exit 1
+    }
+    git checkout -B "$SDK_BRANCH" "fork/$SDK_BRANCH"
```
fantapop: can you use long opts for these?  I see `-b` is being used down below as well.

fantapop (commented):
Left some more feedback here.  Is there anything useful to add tests for?

scripts/openapi-sync.sh
linhcrl: apparently, -b and -B don't have long form

scripts/openapi-sync.sh
linhcrl: made this change

scripts/openapi-sync.sh
linhcrl: updated

scripts/openapi-sync.sh
linhcrl: updated

scripts/invoke-claude-changelog.sh
linhcrl: As discussed I'll run a gap analysis (separate from this PR)

scripts/invoke-claude-changelog.sh
linhcrl: fixed this

CLAUDE.md
linhcrl: I think we should keep these guidelines in the repos' respective CLAUDE.md files. Like you said, having to check out shared skills can be annoying. And this way users that are manually updating the SDK and using Claude to write the CHANGELOG entry can also benefit from having these guidelines, not just the automated workflows. Additionally, the examples are repo specific.

.github/workflows/openapi-sync.yml
linhcrl: done

linhcrl:
> Left some more feedback here. Is there anything useful to add tests for?

added some tests

scripts/invoke-claude-changelog.sh
linhcrl: I'm no longer dynamically updating the file

CLAUDE.md
fantapop: It seems like the instructions are complex and it's difficult to get claude to follow them. It sucks to have to copy those everywhere.  I'm not concerned about it for this PR though.  We can consider further in the future.

scripts/invoke-claude-changelog.sh
fantapop: I ran one.  It looks like the main gap is the ability to update an existing PR.  Maybe we can consider in the future.

.claude/skills/generate-sync-metadata.md
```diff
@@ -0,0 +1,77 @@
...
+- Titles must be under 72 characters (strict limit)
+- If changes are focused on one specific area, describe that clearly (e.g., "Add MFA audit log actions", "Update invoice field descriptions")
+- If there are multiple UNRELATED changes (2 or more distinct user-facing changes), you have TWO valid options:
+  1. Mention ALL changes clearly in the title (must be under 72 chars AND clear)
+  2. Use the generic title: "Sync OpenAPI spec from managed-service"
+- DO NOT cherry-pick - do not mention only some changes while omitting others
+- When in doubt or if the title would exceed 72 characters, use the generic title
+- Clarity is more important than brevity - if you can't be both clear AND concise, use the generic title
+
+**Example of what NOT to do:** Set title to "Add CreateApiKey endpoint and Region.S3VpcEndpointId field" when there are also DeleteBlackoutWindow, UpdateBackupConfiguration, ListInvoices endpoints and Invoice.Adjustments, Invoice.Status field changes.
```
fantapop: It kind of seems like this is overly specific.  Maybe something like:

** read the title you've chosen again and the list of endpoints affected.  If the list of endpoints contains values not listed in the title, use the generic title and list the endpoints in the commit message description.

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,59 @@
...
+      MANAGED_SERVICE_TOKEN: ${{ secrets.MANAGED_SERVICE_TOKEN }}
+      FORK_OWNER: crl-gh-actions-pr-bot
+      FORK_PUSH_TOKEN: ${{ secrets.FORK_PUSH_TOKEN }}
+      CREATE_PR_TOKEN: ${{ github.token }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          token: ${{ github.token }}
```
fantapop: nit: I'm surprised this isn't the default.  I don't recall ever setting token for this action.

scripts/git-askpass.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# GIT_ASKPASS script for fork authentication. Reads credentials from
+# environment variables so the token is never written to disk.
+set -euo pipefail
+
+# Get the script directory to find helper functions
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source "$SCRIPT_DIR/lib/actions-helpers.sh"
+
+if [ -z "$GIT_FORK_USER" ] || [ -z "$GIT_FORK_PASSWORD" ]; then
```
fantapop: nit:  In autosolve, I changed this in the latest PR to not use the fork nomenclature.  It seemed unnecessary since it's really just a generic way to pass user and password to this script.   Here is the full file if you want to copy it:

```
#!/bin/sh
# GIT_ASKPASS script for fork authentication. Reads credentials from
# environment variables so the token is never written to disk.
set -e

if [ -z "$GIT_USER" ] || [ -z "$GIT_PASSWORD" ]; then
  echo "error: GIT_USER and GIT_PASSWORD must be set" >&2
  exit 1
fi

case "$1" in
  Username*) echo "$GIT_USER" ;;
  Password*) echo "$GIT_PASSWORD" ;;
esac
```

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+
+  if [[ -z "${managed_service_pr_url:-}" ]]; then
+    log_error "managed_service_pr_url is required"
+    return 1
+  fi
+
+  if [[ -z "${CREATE_PR_TOKEN:-}" ]]; then
+    log_error "CREATE_PR_TOKEN environment variable is required"
+    return 1
+  fi
```
fantapop: nit: It seems like from within this script it should just be a generic name if the script only needs one token.  i.e. GIT_TOKEN or GH_TOKEN or GITHUB_TOKEN, etc..

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+    exit 1
+  fi
+
+  if [[ "$EVENT_TYPE" == "openapi-spec-merged" && -z "${MANAGED_SERVICE_SHA:-}" ]]; then
+    log_error "MANAGED_SERVICE_SHA is required for openapi-spec-merged event"
+    exit 1
+  fi
+
+  if [[ "$EVENT_TYPE" == "openapi-spec-changed" && -n "${MANAGED_SERVICE_SHA:-}" ]]; then
+    log_error "Unexpected parameter for openapi-spec-changed event: MANAGED_SERVICE_SHA"
```
fantapop: nit: this message is slightly confusing.  should it have a $ in front?

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+  # that the fork doesn't have. This prevents feature branches from appearing to add
+  # files requiring special permissions (e.g., workflows), which would cause permission issues
+  log_info "Syncing fork's $default_branch with origin/$default_branch"
+
+  # Set up askpass for fork push
+  export GIT_ASKPASS="$SCRIPT_DIR/git-askpass.sh"
+  export GIT_FORK_USER="$FORK_OWNER"
+  export GIT_FORK_PASSWORD="$FORK_PUSH_TOKEN"
+  export GIT_TERMINAL_PROMPT=0
+  chmod +x "$GIT_ASKPASS"
```
fantapop: nit: can we just set +x on the file as committed?

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+    TIMESTAMP=$(date --utc +%Y%m%d-%H%M%S)
+    HEAD_BRANCH="automation/openapi-spec-change-$TIMESTAMP"
+    log_info "Creating new head branch: $HEAD_BRANCH from $BASE_BRANCH"
+
+    # Set up askpass for origin operations
+    export GIT_ASKPASS="$SCRIPT_DIR/git-askpass.sh"
+    export GIT_FORK_USER="x-access-token"
+    export GIT_FORK_PASSWORD="$CREATE_PR_TOKEN"
+    export GIT_TERMINAL_PROMPT=0
+    chmod +x "$GIT_ASKPASS"
```
fantapop: nit: I've seen this a few times now. Is it possible to bundle it up into a function like
```
setGitAuth "x-access-token" "$CREATE_PR_TOKEN"
```

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+    log_info "Fetching openapi.json from $MS_OWNER/$MS_REPO PR #$MS_PR_NUMBER"
+  fi
+
+  # Fetch the OpenAPI spec file from managed-service
+  export GH_TOKEN="$MANAGED_SERVICE_TOKEN"
+  gh api "repos/$MS_OWNER/$MS_REPO/contents/$SPEC_PATH_IN_MS?ref=$REF" \
+    --jq '.content' | base64 --decode > "$LOCAL_SPEC_PATH.new"
+
+  # Validate the downloaded file is valid JSON
+  if ! jq empty "$LOCAL_SPEC_PATH.new" 2>/dev/null; then
```
fantapop: do we have to redirect stderr?  It seems like it would be handy to see if there was a failure

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+    log_error "Downloaded spec file is not valid JSON"
+    rm "$LOCAL_SPEC_PATH.new"
```
fantapop: nit: I've seen this path with `.new` at the endpoint multiple times. How about pull to a variable?  i.e. $LOCAL_SPEC_PATH_NEW

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,548 @@
...
+
+# Create or update PR
+# Creates a new SDK PR targeting the base branch or updates an existing PR's
+# title and description with the latest from Claude.
+create_or_update_pr() {
+  if [[ "$HAS_CHANGES" != "true" ]]; then
+    return
+  fi
+
+  export GH_TOKEN="$CREATE_PR_TOKEN"
```
fantapop: nit: could set GH_TOKEN outside of the function.

fantapop (approved):
I left a bunch of nits but these all seem pretty surface level so I'm  going to provide sign off.

scripts/lib/openapi-sync-helpers.sh
linhcrl: nope I was referring to the parameter name. But I updated it to `Input 'sha' should not be provided for openapi-spec-changed event` instead because it's more helpful to refer to the github workflow input name, which is just `sha`

scripts/git-askpass.sh
linhcrl: makes sense. updated.

scripts/lib/openapi-sync-helpers.sh
linhcrl: good idea, done

scripts/lib/openapi-sync-helpers.sh
linhcrl: ah good point. Updated

scripts/lib/openapi-sync-helpers.sh
linhcrl: updated to GH_TOKEN

.github/workflows/openapi-sync.yml
linhcrl: removed, this wasn't necessary

.claude/skills/generate-sync-metadata.md
linhcrl: updated with `For multiple unrelated changes: review your title against all affected endpoints and fields. If your title doesn't cover everything or exceeds 72 characters, use the generic title "Sync OpenAPI spec from managed-service" and describe the changes in the commit body instead.`

scripts/lib/openapi-sync-helpers.sh
linhcrl: removed redirect

scripts/lib/openapi-sync-helpers.sh
linhcrl: updated

scripts/lib/openapi-sync-helpers.sh
linhcrl: this is (re)moved now

.github/workflows/openapi-sync.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
```
fantapop: what uses go in this?

scripts/lib/openapi-sync-helpers.sh
```diff
@@ -0,0 +1,514 @@
...
+    return 1
+  fi
+
+  # Create temp files
+  local prompt_file=$(mktemp)
+  local output_file=$(mktemp)
+  trap "rm -f \"$prompt_file\" \"$output_file\"" EXIT RETURN
+
+  # Copy prompt template to temp file
+  cp "$prompt_template" "$prompt_file"
```
fantapop: Did we talk about this already? if we're no longer editing the prompt can we get rid of the copy and "prompt template" naming here?

fantapop (commented):
I left a couple more questions inline.  And also here are some notes about security hardening for you to consider:

* the system prompt is static and we're not templating anymore so there's no chance of prompt injection at least into the system prompt.  👍 
* The prompt is being loaded from a local file, so maybe a PR could submit a change to that file and have it run if we were to run the action on the branch? This is not a risk as is, but maybe a future change allows that without thinking about it?  is there a way we can lock this to only running on certain branches? i.e. main   Or do we also need the pending deploy branches?  ⚠️ 
* Currently the action only runs on workflow dispatch. You need write access to the repo to be able to trigger that so if you had write access, you wouldn't need to run the action to compromise anything. 👍 
* The prompt is telling claude to look at code already in the sdk so we don't really need to worry about prompt injection instructions being inside the code.  If they were, the exploit would have already happened 👍 
* removing the unnecessary env vars from the claude call is helpful but remains a point of sync for those env vars.  If someone changed the name they would get through.  A better model would be to use an allow list of env vars instead of a deny list.     Another good option that would require more code changes would be to divide openapi-sync.sh script up in to smaller scripts and call them in individual steps.  You could then supply the env vars that are needed just for that step.  Currently all steps are getting all the env vars.  If there was a supply change compromise in any of the google actions they would all get the github tokens. ⚠️

.github/workflows/openapi-sync.yml
linhcrl: Added a comment explaining Go is required by make generate-openapi-client and make validate

scripts/lib/openapi-sync-helpers.sh
linhcrl: Fixed - removed the temp file copy and renamed to prompt_file. Now reads the file directly.

linhcrl:
> I left a couple more questions inline. And also here are some notes about security hardening for you to consider:
> 
> * the system prompt is static and we're not templating anymore so there's no chance of prompt injection at least into the system prompt.  👍
> * The prompt is being loaded from a local file, so maybe a PR could submit a change to that file and have it run if we were to run the action on the branch? This is not a risk as is, but maybe a future change allows that without thinking about it?  is there a way we can lock this to only running on certain branches? i.e. main   Or do we also need the pending deploy branches?  ⚠️
> * Currently the action only runs on workflow dispatch. You need write access to the repo to be able to trigger that so if you had write access, you wouldn't need to run the action to compromise anything. 👍
> * The prompt is telling claude to look at code already in the sdk so we don't really need to worry about prompt injection instructions being inside the code.  If they were, the exploit would have already happened 👍
> * removing the unnecessary env vars from the claude call is helpful but remains a point of sync for those env vars.  If someone changed the name they would get through.  A better model would be to use an allow list of env vars instead of a deny list.     Another good option that would require more code changes would be to divide openapi-sync.sh script up in to smaller scripts and call them in individual steps.  You could then supply the env vars that are needed just for that step.  Currently all steps are getting all the env vars.  If there was a supply change compromise in any of the google actions they would all get the github tokens. ⚠️

Thanks for the review. I made changes to address these points: 

**Branch restriction**: Added a validation step that checks the workflow is only running on main branch (lines 39-45). This prevents  PRs from modifying the prompt file and executing it.

**Env var isolation**: Split the workflow into separate steps with scoped environment variables:
- "Generate updated client" step gets the tokens needed for fetching the spec and regenerating code
- "Update changelog with Claude" step only receives BASE_BRANCH and HAS_CHANGES - no secrets are exposed to Claude 
- "Publish changes" step gets only the tokens needed for git operations

This reduces the blast radius if any step or dependency is compromised

fantapop (approved):
nice. this is looking good to me now

linhcrl:
> nice. this is looking good to me now

Thanks @fantapop ! I will mention some of these discussion/decisions in the documentation I'm working on

PR #109, opened 2026-05-18
Fix release workflow to use workflow_dispatch for ccloud-private

linhcrl:
The ccloud-private repo's sync-cli-with-sdk workflow uses workflow_dispatch, not repository_dispatch. This updates the release workflow to trigger it correctly using `gh workflow run` with the required sdk_version input.

Also fixes all references from tag-created to tag_created for consistency with the autotag-from-changelog action's output format.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     outputs:
-      tag-created: ${{ steps.autotag.outputs.tag-created }}
+      tag_created: ${{ steps.autotag.outputs.tag_created }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     outputs:
-      tag-created: ${{ steps.autotag.outputs.tag-created }}
+      tag_created: ${{ steps.autotag.outputs.tag_created }}
```
linhcrl, line 15: the reusable workflow outputs `tag_created` in snake case, not kebab cases

fantapop (approved):
LGTM

PR #112, opened 2026-05-28
[CNSL-2286] Add managed-service PR author as reviewer on SDK sync PRs

linhcrl:
When the OpenAPI sync workflow opens a new SDK PR, look up the managed-service PR's author and request them as a reviewer. The lookup runs in the generate step (which already has the managed-service token) and is exposed as a step output; the publish step adds the reviewer only on the create-new-PR path, never when updating an existing PR.

Reviewer assignment is best-effort: if the lookup returns empty or the gh API call rejects the user, the workflow logs a warning and continues rather than failing the sync.

### DIFF
--- a/scripts/lib/openapi-sync-helpers.sh
+++ b/scripts/lib/openapi-sync-helpers.sh
@@ -93,12 +93,13 @@ validate_event_payload() {
 # Parse managed-service PR metadata from URL
 #
 # Validates that the URL is from cockroachlabs/managed-service and extracts
-# the PR number.
+# the PR number and author.
 #
 # Exports:
 #   MS_OWNER: The GitHub organization
 #   MS_REPO: The repository name
 #   MS_PR_NUMBER: The managed-service PR number
+#   MS_PR_AUTHOR: The managed-service PR author's GitHub login (empty on lookup failure)
 parse_managed_service_pr_url() {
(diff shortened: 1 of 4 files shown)


scripts/lib/openapi-sync-helpers.sh
```diff
@@ -113,8 +114,18 @@ parse_managed_service_pr_url() {
 
   log_info "Managed-service PR: $MS_OWNER/$MS_REPO#$MS_PR_NUMBER"
 
+  # Look up PR author so we can request them as a reviewer on the SDK PR.
+  # Best-effort: failures here must not abort the sync.
+  MS_PR_AUTHOR=$(GH_TOKEN="$MANAGED_SERVICE_TOKEN" \
+    gh api "repos/$MS_OWNER/$MS_REPO/pulls/$MS_PR_NUMBER" --jq '.user.login' 2>/dev/null || true)
```
fantapop: It seems okay to not fail here but it would be nice to let standard err escape so we can see what happened.

fantapop (approved):
I left one nit.  Thanks for making this change!

scripts/lib/openapi-sync-helpers.sh
linhcrl: Updated! TFTR

PR #114, opened 2026-06-03
[CNSL-2288] Add automated fork main sync with upstream

linhcrl:
Adds sync-fork.yml: on every push to upstream main (and via manual workflow_dispatch), calls the cockroachdb/actions sync-fork reusable workflow to push upstream's main to crl-gh-actions-pr-bot's fork's main. 

The reusable workflow no-ops unless github.repository matches the configured upstream_repo, so the file is safe to land on both upstream and any fork.

### DIFF
--- a/.github/workflows/sync-fork.yml
+++ b/.github/workflows/sync-fork.yml
@@ -1,0 +1,23 @@
+name: Sync Fork
+
+# Pushes cockroachdb/cockroach-cloud-sdk-go's main to the fork's main on every
+# commit to main. The reusable workflow no-ops when invoked anywhere other
+# than the configured upstream_repo.
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  sync:
+    uses: cockroachdb/actions/.github/workflows/sync-fork.yml@v0
+    with:
+      upstream_repo: cockroachdb/cockroach-cloud-sdk-go
+      fork_repo: crl-gh-actions-pr-bot/cockroach-cloud-sdk-go
+      allow_fork_force_sync: false
+    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-fork.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  sync:
+    uses: cockroachdb/actions/.github/workflows/sync-fork.yml@v0
+    with:
+      upstream_repo: cockroachdb/cockroach-cloud-sdk-go
+      fork_repo: crl-gh-actions-pr-bot/cockroach-cloud-sdk-go
+      allow_fork_force_sync: true
+    secrets:
+      fork_push_token: ${{ secrets.FORK_PUSH_TOKEN }}
```
fantapop: since this token will require more permissions, perhaps we should name the secret specifically for this action?  I was thinking something like SYNC_FORK_TOKEN

.github/workflows/sync-fork.yml
```diff
@@ -0,0 +1,23 @@
...
+permissions:
+  contents: read  # Reusable workflow uses GITHUB_TOKEN to fetch this repo's main
+
+jobs:
+  sync:
+    uses: cockroachdb/actions/.github/workflows/sync-fork.yml@v0
+    with:
+      upstream_repo: cockroachdb/cockroach-cloud-sdk-go
+      fork_repo: crl-gh-actions-pr-bot/cockroach-cloud-sdk-go
+      allow_fork_force_sync: true
```
fantapop: do we need force here?  The action calls that out as something that should be for testing only.

linhcrl: updated

fantapop (approved):
LGTM

PR #117, opened 2026-06-25
Stream Claude CLI events live in OpenAPI sync workflow log

linhcrl:
Switch the Claude CLI from --output-format json to stream-json so it emits one JSON event per turn instead of buffering a single object at the end, and pipe through tee so Claude's events stream to the GitHub Actions log live while still being captured to the output file. Wrap the stream in a ::group:: section so the per-turn JSON collapses in the workflow UI.

Update the jq extractor to select the final type:"result" event from the Claude output since the captured file is now newline-delimited events instead of one object.


fantapop (approved):
LGTM

PR #118, opened 2026-06-29
Sync OpenAPI spec and regenerate client

mwstobo:
Regenerate the SDK client from the latest cockroachlabs.cloud OpenAPI spec.

Added:
- New audit log actions for MFA management and Migration Assistant lifecycle, plus new audit log sources (CC API, CLI, UI, internal, MCP, SSH gateway).
- SystemActorName on audit log entries.
- RegionMachineSpecs on dedicated cluster create/update specs and DiskIops, MachineType, NumVirtualCpus on the Region model.
- EnableSendingQueue option on log export configuration.
- STATUS_STALE private endpoint connection status.

Changed (breaking):
- Request body types renamed from *Request to *Body (e.g. CreateSQLUserRequest -> CreateSQLUserBody, UpdateDatabaseRequest -> EditDatabaseBody, GetGroupRequest -> GetGroup2Body).
- `machine_spec` is now optional on `DedicatedHardwareCreateSpecification` (exactly one of `machine_spec` or `region_machine_specs` must be provided). `MachineSpec` is now a pointer and NewDedicatedHardwareCreateSpecification` no longer takes a `machineSpec` argument.


mwstobo:
cc @fantapop, as this includes some of your breaking changes as well!

linhcrl (commented):
`DedicatedHardwareCreateSpecification.MachineSpec` went from required to optional (value → pointer). That's source-breaking for callers, so probably deserves its own Breaking Change bullet. 

This looks good to me otherwise, but I would get sign-off from @fantapop because I'm not sure what his release plan is for the changes he's working on

fantapop (commented):
In order to rev the major there are a few changes throughout the repo we should make.  @linhcrl did some of these get automated I forget.   You can look at this sha where 7.0.0 was created: 1c965449b043979ae8afed610be04051e8f37267

For example the README needs to change and internal/spec/config.yaml needs to change and pkg/client/configuration.go needs to change.   I think that's it...

Also, it would be nice if this was divided into 2 commits since the one from my grpc-gateway version change is so large.  It's making the new one hard to review.

CHANGELOG.md
```diff
@@ -32,13 +32,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  cluster create/update specifications and `DiskIops`, `MachineType`, and `NumVirtualCpus` on the
+  `Region` model.
 
 ### Changed
 
 - Update the OpenAPI sync workflow to request the managed-service PR author as a
   reviewer on the generated SDK PR.
 - Updated release workflow to trigger ccloud-private CLI sync using workflow_dispatch
   instead of repository_dispatch.
+- Breaking Change: `machine_spec` is now optional on `DedicatedHardwareCreateSpecification`
```
fantapop, line 45: nit: newer entries should be put in top down order like a stack

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.0.0] - 2026-07-06
```
fantapop, line 10: ideally this gets added in your new commit for bumping the version.

fantapop (approved):
I left some minor feedback items.  It's all mostly nits.  One other meta feedback is that the prefixing of the pr name into the commit titles seems unnecessary.  It would be clearer to be more specific or have shorter commit titles rather than trying to group them by a prefix which is already pretty generic and overused IMO.

CHANGELOG.md
mwstobo: Done!

CHANGELOG.md
mwstobo: Done!

PR #119, opened 2026-06-30
Allow sed -n in OpenAPI sync Claude allowlist

linhcrl:
Large diffs get truncated to tool-result files; Claude needs read-only sed to extract sections, otherwise it burns turns on denied commands and hits max-turns.


fantapop (approved):
Lgtm

PR #120, opened 2026-07-09
Pre-generate git diff for Claude changelog step in OpenAPI sync

linhcrl:
The changelog step failed on every recent run: Claude ran git diff itself, the huge generated-client diff spilled into an oversized tool result, and Claude burned its 15 turns on shell pipelines that the allowlist rejected before it could edit CHANGELOG.md.

Generate the diff (plus a --stat summary) in the script and have Claude read the files directly instead. This removes the need for any Bash access, tightening the allowlist to Read and Edit(CHANGELOG.md), and raises --max-turns to 30 for headroom on large syncs.


fantapop (approved):
LGTM

PR #130, opened 2026-08-03
Sync OpenAPI spec to deployed API (v9.0.0)

fantapop:
Brings the SDK up to the deployed CockroachDB Cloud API (v9.0.0), rebuilt **commit-by-commit to mirror every managed-service commit that changed the published OpenAPI spec** since v8.0.0 — not just the ones that happened to get an SDK sync PR. Each commit is regenerated from the spec at that managed-service commit (`Managed-service-commit-SHA` trailer), so the result matches the deployed spec exactly (0 diff vs `cockroachlabs.cloud/.../latest/openapi.json`). `go build`, `go vet`, and `make validate` pass, and a full regenerate is zero-diff.

Supersedes the stuck sync PRs #123/#125/#128 (and #126, whose StorageType field is `DEVELOPMENT`-only visibility and correctly not in the published spec, so it's excluded).

### Breaking change (drives the major bump)
- `AuthPrincipal` is no longer required for `EnableLogExport`. `NewEnableLogExportBody` no longer takes `authPrincipal`, and `AuthPrincipal` is now a `*string`. (Required→optional in the spec is a relaxation for REST callers, but changes the generated Go constructor — hence 9.0.0.)

### Added
- OTLP/HTTP log export (`OTLP_HTTP`) with `OtlpEndpoint`/`OtlpHeaders`, plus delivery health (`DeliveryStatus`, `DeliveryStatusMessage`).
- Azure Log Analytics v2 log export (`AZURE_LOG_ANALYTICS_V2`).
- Per-cluster runtime scanning operations (`EnableClusterRuntimeScanning`, `GetClusterRuntimeScanning`, `DisableClusterRuntimeScanning`).
- `service_account_name` and `roles` filter query parameters on ListServiceAccounts.
- `UPDATE_BACKUP_CONFIGURATION` audit log action; role-management group support (docs); CLUSTER_DEVELOPER read access to private-endpoint operations.

### Deprecated
- `azure_shared_key` on log export configuration (superseded by the Azure Log Analytics v2 Data Collection Rule fields).

### Also
- Fixes a latent bug in the Go service template (optional array query parameters referenced the bare param name instead of the options-struct field) — required to compile the new `roles` filter.

### Commits
1. Fix optional array query parameter generation in the Go service template
2. role management groups + CLUSTER_DEVELOPER endpoint access (ms `4620dfbd`)
3. ListServiceAccounts name/role filters (ms `ead24b08`)
4. per-cluster runtime scanning API (ms `852931a9`)
5. Azure Log Analytics v2 log export (ms `32e76227`) — breaking; deprecates `azure_shared_key`
6. OTLP/HTTP log export sink (ms `5caf5243`)
7. rename ListServiceAccounts filters (ms `58e8fc0f`)
8. ClusterStatus.Type enum doc comments (ms `6ff41c84`)
9. Tag for 9.0.0 (version bump)

Changelog entries are under `[Unreleased]`; the release is not git-tagged here (tagging happens after this PR merges).

🤖


fantapop:
One slight downside of this approach is that it pulls in churn that wouldn't otherwise be necessary.  For example, we have a commit for adding the new ListServiceAccounts params and then one later for renaming them.  While it does show unnecessary churn, I think there is some merit of leaving it in.  It certainly will simplify the process once the automation is flowing.

CHANGELOG.md
```diff
@@ -9,15 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Changed
 
+- Breaking Change: `AuthPrincipal` is no longer required for `EnableLogExport`. `NewEnableLogExportBody` no longer takes `authPrincipal`, and `AuthPrincipal` is now a pointer.
 - Update role management operation descriptions to reflect support for groups in addition to users and service accounts.
 - Grant `CLUSTER_DEVELOPER` role read access to private endpoint operations (ListAwsEndpointConnections, ListEgressPrivateEndpoints, GetEgressPrivateEndpoint, ListPrivateEndpointConnections, ListPrivateEndpointServices).
 
+### Deprecated
+
+- Deprecate `azure_shared_key` on log export configuration; use the Azure Log Analytics v2 (Data Collection Rule) fields instead.
+
```
linhcrl, line 30: We might also want to add the following for completeness 

```- Deprecate the `AZURE_LOG_ANALYTICS` log export type; use `AZURE_LOG_ANALYTICS_V2` instead.```

linhcrl: Might also be worth mentioning it in the commit message similar to `azure_shared_key`

linhcrl (approved):
Approving with one small suggestion

CHANGELOG.md
fantapop: Good catch — added it. The `Deprecated` section now includes:

`- Deprecate the AZURE_LOG_ANALYTICS log export type; use AZURE_LOG_ANALYTICS_V2 instead.`

I also amended the "Azure Log Analytics v2 log export" sync commit message to mention the type deprecation alongside `azure_shared_key`. (Force-pushed, so the commit SHAs from that point on have changed.) 🤖

PR #131, opened 2026-08-07
[CNSL-2477] Automerge automated SDK PRs and request review

linhcrl:
Add automerge_sdk_pr, called from the publish step for openapi-spec-merged events only, which rebase-merges the SDK PR into its pending deploy branch once the managed-service PR has landed.

Rebasing lands the PR's commits as they are, preserving the Managed-service-commit-SHA trailer that pending-deploy-check reads.

GitHub recomputes mergeability asynchronously and the publish step force-pushes moments earlier, so an upfront mergeability check would only ever read a stale answer. The merge is attempted directly and retried instead, stopping early on a conflict once GitHub reports a state for the commit we pushed.

In case of a merge failure, the workflow tags console team members.

This PR also contains updates to the pending deploy workflow to ensure console team members are tagged as reviewers when opening PRs to land changes from a pending deploy branch on main.

------

Tested a clean case and it merged as expected. Tested on a PR with a conflict and it posted the comment as expected:
[
Screenshot 2026-08-06 at 11 26 07 PM
](url)

### DIFF
--- a/scripts/lib/openapi-sync-helpers.sh
+++ b/scripts/lib/openapi-sync-helpers.sh
@@ -567,3 +567,120 @@ create_or_update_pr() {
     fi
   fi
 }
+
+# Post a comment on the SDK PR explaining why automerge could not complete.
+#
+# Args: $1 - human-readable reason
+post_automerge_failure_comment() {
+  local reason="$1"
+
+  # Never let a comment failure mask the automerge failure it is reporting: this script runs
+  # under `set -e` and the caller still needs to exit non-zero.
+  gh pr comment "$SDK_PR_NUMBER" --repo "$GITHUB_REPOSITORY" --body "$(cat <<EOF
+## Automerge Failed
+
+The managed-service PR has merged and this PR was updated, but it could not be merged into
+\`$BASE_BRANCH\` automatically:
+
+> $reason
+
+Resolve this manually and merge the PR.
+EOF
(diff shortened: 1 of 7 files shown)


scripts/lib/openapi-sync-helpers.sh
```diff
@@ -567,3 +567,119 @@ create_or_update_pr() {
...
+    fi
+
+    log_info "Merge attempt $attempt/$max_attempts failed: $merge_output"
+    sleep "$retry_seconds"
+  done
+
+  log_error "Could not merge SDK PR #$SDK_PR_NUMBER into $BASE_BRANCH: $merge_output"
+  post_automerge_failure_comment "$(describe_unmergeable_pr)"
+  return 1
+}
```
fantapop, line 686: Will it be obvious if these are failing?  Maybe we should add one of use to a pr that fails to merge.

fantapop (approved):
I'm generally good with this. I'm still unclear how we'll start to see failures though or know when we're supposed to land the integration branch. I left one comment.

scripts/lib/openapi-sync-helpers.sh
linhcrl: Yep it posts a comment on the failing PR and I updated the PR to also add both of us as reviewers when it fails

linhcrl: Additionally, I updated the workflow that opens a PR to land the changes on the pending deploy branch in main

PR #132, opened 2026-08-10
Rotate pending deploy branch by git ancestry instead of timestamps

linhcrl:
Comparing timestamps parsed from branch names only approximates whether the live pending deploy branch is stale, and depends on finding the last merged pending-deploy PR within the 50 most recent merged PRs. When that search comes up empty the job reuses a branch whose contents have already shipped, leaving open PRs stacked on a base behind main.

Ask git directly instead: if the branch is an ancestor of main, it is spent and a fresh one is cut.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,11 +42,10 @@ jobs:
 
   # After a successful release, this job ensures there's always a fresh pending deploy
   # branch ready for the next release cycle. It:
-  # 1. Finds the most recently merged pending deploy branch (from PRs to main)
-  # 2. Finds any existing live pending deploy branches
-  # 3. Creates a new pending deploy branch if there are no live ones or if the live
-  #    ones are stale (associated with an older date than the last merged one)
-  # 4. Retargets all open PRs to point to the current pending deploy branch
+  # 1. Finds any existing live pending deploy branches
+  # 2. Creates a new pending deploy branch if there are no live ones, or if the live
+  #    one is already fully contained in main (meaning its contents have shipped)
+  # 3. Retargets all open PRs to point to the current pending deploy branch
   manage-pending-deploy:
     needs: release
     if: needs.release.outputs.tag_created == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,11 +42,10 @@ jobs:
...
   # branch ready for the next release cycle. It:
-  # 1. Finds the most recently merged pending deploy branch (from PRs to main)
-  # 2. Finds any existing live pending deploy branches
-  # 3. Creates a new pending deploy branch if there are no live ones or if the live
-  #    ones are stale (associated with an older date than the last merged one)
-  # 4. Retargets all open PRs to point to the current pending deploy branch
+  # 1. Finds any existing live pending deploy branches
+  # 2. Creates a new pending deploy branch if there are no live ones, or if the live
+  #    one is already fully contained in main (meaning its contents have shipped)
+  # 3. Retargets all open PRs to point to the current pending deploy branch
```
fantapop, line 48: It seems like these numbers are still off from the numbers listed below.  originally the ones below went up to 5 so it was off previously.  Now the ones below go up to 4...

.github/workflows/release.yml
```diff
@@ -108,49 +73,35 @@ jobs:
...
           if [ -z "$CURRENT_BRANCH" ]; then
-            # No current live branch exists, create new
             log_info "No current live branch exists, creating new branch..."
             selected_branch="automation/pending-deploy-${current_timestamp}"
             needs_creation=true
-          elif [ -n "$PREVIOUS_BRANCH_TIMESTAMP" ] && [ "$PREVIOUS_BRANCH_TIMESTAMP" -ge "$CURRENT_BRANCH_TIMESTAMP" ]; then
-            # Previous timestamp >= current timestamp, create new branch
-            log_info "Previous timestamp ($PREVIOUS_BRANCH_TIMESTAMP) >= current timestamp ($CURRENT_BRANCH_TIMESTAMP), creating new branch..."
+          elif git merge-base --is-ancestor "origin/$CURRENT_BRANCH" origin/main; then
+            log_info "$CURRENT_BRANCH is fully merged into main, creating new branch..."
```
fantapop, line 100: I'm not sure I'm fully understanding whats happening here.  Is this happening when we tag a release?  So the workflow would be:

1. prs from manage service open prs against the current pending deploy branch
2. a merged managed-service pr automerges in the current pending deploy branch
3. on release of managed-service a PR to merge current pending deploy branch into main is created
4. we manually review that
5. on merge of that (I'm presuming that PR will include a tag), a release is kicked off and this will run?

What happens if a PR from managed-service lands just after a deploy but before we've add a chance to review the SDK PR of the release?

linhcrl: Steps 1-4 are correct. 5 isn't: merging the pending deploy PR kicks off `create-release-pr.yml`, which opens a version bump PR. That also needs a manual approve and merge. Merging that is what kicks off the release workflow, and this job runs at the end of it.

_About the race_: It won't slip in unnoticed. Every generated SDK commit carries a trailer naming the managed-service commit it came from, and a CI check on the pending deploy PR walks each commit and confirms that managed-service commit is released. GitHub re-runs it whenever new commits land on the branch, so a late arrival turns the PR red and posts a comment listing which commits aren't deployed yet.

It's not a hard block though - that check isn't in main's required status checks (only license/cla is), so nothing actually stops someone from merging past it. Happy to make it required.

I considered cutting a second pending deploy branch at deploy time to freeze the in-flight one, but the automation assumes exactly one live branch at a time - two would mean stacked branches with merge-ordering constraints. I didn't think it was worth the complexity for something that resolves itself on the next deploy.

fantapop: Okay, so a late arrival would necessitate some manual fixing of the pending deploy branch?  In that case are the steps:
1. remove the commit of non-deployed change from the pending deploy branch (I assume it would build again automatically)
2. merge the pending deploy branch
3. manually create a new pending deploy branch (or does one of these steps automatically do that?)
4. manually push the commit from step 1 on to the new branch

It seems like we could add the release tag as part of the pending deploy branch creation if we're planning to keep the review there.  But I'm fine with waiting a couple cycles to see how it goes.

fantapop (approved):
LGTM

